PR #6, opened 2016-02-25
Fix checksums missed in Firefox 38.6.1esr update.

WarheadsSE:
As the title says. A little embarrassed that I missed this.


shr-project:
This actually breaks the checksums

firefox-l10n-pl: Check if the license information has changed in /home/jenkins/oe/world/shr-core/openembedded-core/LICENSE to verify that the LICENSE value "MPLv1 | GPLv2+ | LGPLv2.1+" remains valid [license-checksum]

Are you building it with some very old oe-core? IIRC the LICENSE files were updated before Dizzy release.

WarheadsSE:
Yeah, this one is my bad. It needs reverted on master.

Jason Plum
WarheadsSE

otavio:
@shr-project as you have commit, can you handle the revert please? I am under heavy load here.

shr-project:
I've already reverted it (Just after WarheadsSE's message).

PR #10, opened 2016-06-15
chromium: Fix build issue with i686

mdavis777:
Previously a build of i686 would fail looking for include.gypi.
With this new directory it can be found by both i686 and i586 targetted
builds.

Signed-off-by: Michael Davis <michael.davis@essvote.com>


mdavis777:
Fixes the issue I was running into where a build of chromium would give me:
WARNING: /opt/yocto/poky/meta-browser/recipes-browser/chromium/chromium_48.0.2548.0.bb: Unable to get checksum for chromium SRC_URI entry include.gypi: file could not be found   
WARNING: /opt/yocto/poky/meta-browser/recipes-browser/chromium/chromium_48.0.2548.0.bb: Unable to get checksum for chromium SRC_URI entry oe-defaults.gypi: file could not be found

PR #13, opened 2016-08-15
Contrib/twoerner/chromium split 3

twoerner:
Building chromium for wayland is a significantly more difficult action than
building it for x11. In order for chromium to run on top of wayland, a shim
is required. This shim is added by checking out a separate repository which
provides patches that are applied to the chromium sources. These patches
are only generated a couple times per year, meaning only a small number of
chromium releases are capable of being run on wayland. This artificially holds
back the versions of chromium which this recipe can provide.

This patch set:
- starts off with a little cleanup (removing an unused patch),
- adds gcc-6 support (making it possible to build successfully with gcc-6),
- splits out a chromium wayland recipe so each can move
  independently and, therefore, lower the maintenance effort while keeping
  up-to-date,
- updates the chromium recipe to 52.0.2743.76 which is considered the
  latest stable (as per https://www.chromium.org/developers/calendar)


twoerner:
Has the fix been pushed?

otavio:
No; no fix has been sent AFAIK.

PR #14, opened 2016-09-02
chromium-wayland: Add back ozone-wayland component

raymanfx:
This was accidentally removed when the Wayland/X11 split happened.


raymanfx:
@twoerner Sorry, I got caught up with other stuff, but finally got around to upload the patch :)

PR #18, opened 2016-12-13
firefox: Remove needless dependency on orbit2 and libidl-native

ashie:
Current firefox doesn't depend on orbit2.
These dependencies originate from too old code:

http://git.openembedded.org/openembedded/commit/?id=c8e5702127e507e82e6f68a4b8c546803accea9d


ashie:
Thanks!

PR #19, opened 2016-12-13
firefox: Don't build system-wide nss & nspr

ashie:
The current mozconfig implies to use buit-in nss & nspr so that
system-wide nss & nspr aren't needed.


otavio:
Instead please change the mozconfig so the system libraries are used.

ashie:
> Instead please change the mozconfig so the system libraries are used.

O.K. I'll close this issue and send another pull request for it.

BTW the following comment exists in the current mozconfig:

```
#ac_add_options --with-system-nss # your version is too old
```

Let me know if we shouldn't ignore it.

FYI: Required version

* nss: 3.21.1
* nspr: 4.12

khrogos or later satisfy these version.

PR #20, opened 2016-12-14
firefox: Use system-wide nss & nspr

ashie:
The recipe requires system-wide nss & nspr but they were not used
because built-in libraries were implicitly enabled. This patch fix
this issue.

Firefox 45ESR requires following versions:

* nss: 3.21.1
* nspr: 4.12

You need krogoth or later to satisfy these version.

### DIFF
--- a/recipes-mozilla/firefox/firefox/mozconfig
+++ b/recipes-mozilla/firefox/firefox/mozconfig
@@ -13,7 +13,8 @@ ac_add_options --disable-strip
 ac_add_options --disable-install-strip
 
 # System libraries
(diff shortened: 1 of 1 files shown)


ashie:
It's the revised version against #19.

ashie:
I noticed that there is one more similar issue. I want to add also `--enable-system-cairo`.
Should I add it to this PR? Or should send another PR?

ashie:
> I've noticed that there is one more similar issue. I want to add also --enable-system-cairo.
> Should I add it to this PR? Or should send another PR?

I'll send another PR for it.

ashie:
Thanks!

PR #21, opened 2016-12-14
firefox: Add some missing items of PRIVATE_LIBS

ashie:
Following libraries aren't specified as PRIVATE_LIBS although they are also built under the private directory:

* liblgpllibs.so
* libmozsqlite3.so
* libbrowsercomps.so
* libclearkey.so"


ashie:
Thanks!

PR #22, opened 2016-12-15
firefox: Use system-wide cairo instead of built-in one

ashie:
Add "--enable-system-cairo" to mozconfig.
Because gtk+ depends on cairo, no need to add it to DEPENDS.


ashie:
Thanks!

PR #33, opened 2017-04-18
Update chromium in morty to match what is in master

andrey-konovalov:
Could you please update chromium in morty to be in sync with current master?

I've cherry picked from master all the commits which changed chromium and which are not in morty.
Tested by building and running on HiKey (aarch64).


otavio:
Sorry but no; this changes the upstream release and breaks some  BSP layers.

andrey-konovalov:
I see. Thanks!

PR #34, opened 2017-04-20
chromium: build ChromeDriver

mrchapp:
This patch enables the build of ChromeDriver. This comes
in very handy for test automation. More information can
be found here:
  https://sites.google.com/a/chromium.org/chromedriver/

Signed-off-by: Daniel Díaz <daniel.diaz@linaro.org>

### DIFF
--- a/recipes-browser/chromium/chromium.inc
+++ b/recipes-browser/chromium/chromium.inc
@@ -51,6 +51,9 @@ do_install() {
 	if [ -f "${B}/out/${CHROMIUM_BUILD_TYPE}/chrome" ]; then
 		install -Dm 0755 ${B}/out/${CHROMIUM_BUILD_TYPE}/chrome ${D}${bindir}/${BPN}/chrome
 	fi
+	if [ -f "${B}/out/${CHROMIUM_BUILD_TYPE}/chromedriver" ]; then
+		install -Dm 0755 ${B}/out/${CHROMIUM_BUILD_TYPE}/chromedriver ${D}${bindir}/${BPN}/chromedriver
+	fi
 	if [ -f "${B}/out/${CHROMIUM_BUILD_TYPE}/icudtl.dat" ]; then
 		install -Dm 0644 ${B}/out/${CHROMIUM_BUILD_TYPE}/icudtl.dat ${D}${bindir}/${BPN}/icudtl.dat
 	fi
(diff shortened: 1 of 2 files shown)


recipes-browser/chromium/chromium.inc
```diff
@@ -87,16 +90,18 @@ do_install() {
...
 FILES_${PN}-codecs-ffmpeg = "${libdir}/${BPN}/libffmpegsumo.so"
 FILES_${PN}-plugin-pdf = "${libdir}/${BPN}/libpdf.so"
+FILES_${PN}-chromedriver = "${bindir}/${BPN}/chromedriver"
 
 FILES_${PN} = "${bindir}/${BPN} ${datadir}/applications ${sbindir}/ ${libdir}/${BPN}/"
 FILES_${PN} += "${bindir} ${libdir}"
 FILES_${PN} += "${bindir}/${BPN}/*.pak"
 FILES_${PN} += "${bindir}/${BPN}/locales/*.pak"
 
+FILES_${PN}-chromedriver-dbg += "${bindir}/${BPN}/.debug/chromedriver"
```
otavio: Please keep all debug together.

twoerner:
By how much does this increase the build time of chromium? If it's a significant amount, it would be better if this were broken out so it is only built for those interested in it specifically.

twoerner:
Or maybe a default-off PACKAGECONFIG?

fboudra:
@mrchapp I prefer if chromedriver is built by default unless the increased build time is significant. In that sense, I do agree with @twoerner . Please share the build time metrics so we can have the full picture.

otavio:
Agreed and also a way to disable it might be on package config as well. Be it enabled or not by default.

twoerner:
Everyone is welcome, of course, to do their own testing. In my testing I found no increase in overall build time. In fact the time to build with the patch went down by almost a minute. Of course I only did one build of each, but suffice to say it's not impacting build time in my one test at all :-)

current master:
    real    47m52.100s
current master + merge request #34:
    real    47m12.625s

mrchapp:
@twoerner, fantastic! Lucky you -- my build only took 5 hours and 10 minutes. I added `ninja chromedriver` as an extra step in do_compile() and that part took 90 seconds. Definitely does not have any significant impact; if anything, it speeds up builds ;)

Would you guys still prefer to add another PACKAGECONFIG for this? Being a separate PACKAGE may suffice, but please advise.

By the way, the .debug part for chromedriver has been fixed, as requested by @otavio, in the original branch (force-pushed, probably messed that one up.)

otavio:
Thanks for handling it. I think that adding a build option is worth the work, specially as it is straightforward.

mrchapp:
Branch refreshed.

PR #36, opened 2017-05-11
chromium-wayland: Fix crash on tab loading

AlexandraSafta:
A SIGILL, Illegal instruction signal was crashing every
chrome tab. It seems we share a common file with qtwebengine, and
the same problem when using a newer Linux was addressed here:

http://code.qt.io/cgit/qt/qtwebengine-chromium.git/commit/?h=49-based&id=b12ffcd411d4776f7120ccecb3be34344d930d2b

Signed-off-by: Tom Hochstein <tom.hochstein@nxp.com>


twoerner:
This patch is already being applied to the non-wayland recipe. If this is also required for wayland, this commit should be reworked to make it common between the two (instead of applying it separately to each recipe individually).

twoerner:
Please see 0003-Remove-hard-coded-values-for-CC-and-CXX.patch as one example.

otavio:
@AlexandraSafta as @twoerner said, please rework it to avoid duplication.

AlexandraSafta:
Thanks for the suggestion. I moved the patch to be applied for both recipes in chromium.inc. 
What do you think about this ? Should I change it to be applied in both recipes, separately ?

twoerner:
Did you mean to say "chromium-browser.inc" instead of "chromium.inc"?

AlexandraSafta:
yes, chromium-browser.inc, sorry. 
I've noticed that 0003-Remove-hard-coded-values-for-CC-and-CXX.patch is applied in both recipes, wayland and non-wayland and this is why I am not sure about my fix.

twoerner:
> yes, chromium-browser.inc, sorry.

Okay, good; just clarifying

> I've noticed that 0003*patch is applied in both recipes

Very true! :-) That's something I should clean up (how ironic!)

I'm just running some builds now. They'll take a couple hours.
Thanks for the contribution.

twoerner:
The patch looks good (sorry for the delay, building chromium a couple times usually takes time).

The last remaining item is the "Signed-off-by:" (SoB) line. It would appear your name is "Alexandra Safta" but the SoB line says "Tom Hochstein"?

AlexandraSafta:
I've put the SoB line with Tom's name because he discover the patch and applied it into our internal layer for fixing the crash. So I considerate fair to be his name on the patch.

otavio:
You can put yours, and his, if he is ok with it.

AlexandraSafta:
Thanks for the suggestion, I have updated the commit message as you suggested.

PR #37, opened 2017-05-12
firefox: Update to 45.9.0esr

ashie:
Also update firefox-l10n to 45.9.0esr

I confirmed it on:

* master
* pyro
* morty


ashie:
We also trying to make a recipe for 52esr but not completed yet.

https://github.com/mozilla-japan/meta-browser/tree/firefox-52.1esr-wip

It includes following patches but we'll remove them when we send a pull request, because they are far from stable:

* EGL support
* Wayland support
* OpenMAX support

PR #38, opened 2017-05-15
chromium: Remove duplicate patch

mdavis777:
Do-not-depend-on-Linux-4.5.patch is a duplicate of unset-madv-free.patch.


otavio:
@mdavis777 it is not a duplication but a revert. Am I missing something?

otavio:
@mdavis777 sorry, my bad. It is indeed a duplication. I will merge it

PR #55, opened 2017-07-19
chromium: remove more Werror flags

yangjiel:
Apply a commit to **not** treat warnings as errors.
(cc1plus: all warnings being treated as errors.)

Original commit from: https://github.com/OSSystems/meta-browser/commit/76f46c4dc91523783a0f209d3f77a7a1404956a8 This commit is the parent commit in last pull request that I submitted.

I have compiled through and without any other error.


kraj:
since its for morty, I dont care as much but I dont want such patches for master. We should
be fixing errors and not hiding them

otavio:
Agreed but it is merged in master already also logically we would bump the release prior making fixes

kraj:
we should try to see what would it take to revert it on head. What are issues we see ?
they should be fixed may be we should do that after we upgrade chromium to latest.

Issue #62, opened 2017-11-10
chromium: Update to a new version


rakuco:
Additional context in [this oe-devel thread](http://lists.openembedded.org/pipermail/openembedded-devel/2017-November/115561.html).

We're currently still packaging Chromium M54 and Chromium M53 + ozone-wayland, both of which are ancient releases by Chromium's standards.

I've [done the work](https://github.com/rakuco/meta-browser/tree/chromium62) to move the existing recipe to M62 (the current stable milestone upstream) based on [my own recipe](https://github.com/rakuco/meta-crosswalk), we now need to break up the ~60 commits in my branch into logical chunks and start merging them.

Depending on how long it takes to get everything merged, we might need to move to M63 directly.

This is a meta issue to track everything we need to do in terms of splitting up the existing commits in my chromium62 branch:
* [x] Drop the ozone-wayland recipe. oz-wl has been unmaintained upstream for years and there is no point in trying to get it to work with M62 or later. Doing so allows us to greatly simplify the rest of the recipe.
* [x] Actual recipe bug fixes that also apply to M54.
* [x] Remove unnecessary patches, as well as patches that can be replaced by gyp options.
* [x] Recipe simplification: drop a few checks for things we build unconditionally, stop setting a few variables here and there, adjust installation directory layout.
* [x] Adjust recipe dependencies: add implicit ones, remove unnecessary ones.
* [x] Move to GN, still on M54. Also requires some additional recipe simplification.
* [x] Move to M62 (should be easy once everything else is done).

rakuco:
@jaragunde @twoerner @kraj

twoerner:
I've don't use wayland, so I'm not interested in ozone-wayland. But if the wayland stuff still works then I'm not opposed to keeping it as its own recipe, stuck at version 53. If it doesn't work then we should get rid of it in master and provide a check in the new chromium6x recipe to make sure the user isn't trying to build against wayland.

rakuco:
I _think_ it currently works in the sense that right now master builds with it, but I've never tried it myself and am not a big fan of the idea of shipping a dangerously old and insecure Chromium version.

All the other changes I've made depend on getting rid of the existing oz-wl recipe to simplify the main recipe and allow us to move to a newer release, so if anything I'd rather have a completely separate recipe (i.e. `recipes-browser/chromium-ozone-wayland`) with a copy of what we have at the moment than having to deal with it in recipes-browser/chromium.

twoerner:
[...]

PR #63, opened 2017-11-13
Remove chromium-wayland recipe and related tasks/settings.

rakuco:
This is a first step towards updating the main Chromium recipe to a more
recent milestone. The ozone-wayland project has been unmaintained for a long
time, and we will not be able to keep supporting it as-is without
significant overhead; meanwhile, there has been work upstream to make
Chromium work on Wayland, and we should be able to leverage it within a few
milestones.

See also: #62.


otavio:
@twoerner I am fine with this change. If you are fine too, please merge.

PR #64, opened 2017-11-13
chromium: Minor, assorted recipe improvements.

rakuco:
This is the second step of many towards moving the existing Chromium recipe to M62.

This pull request is also quite simple even though it has more than 1 commit: mostly expanding `DESCRIPTION` and `HOMEPAGE`, switching to an https fetch URL and a few other things that couldn't fit in other categories of changes we need to land.

See also: #62.


rakuco:
@otavio @twoerner

PR #67, opened 2017-11-13
chromium: Simplify the M54 recipe.

rakuco:
This is one of the bigger chunks of the changes required for us to get to Chromium 62.

This pull request simplifies the Chromium recipe now that the CEF and ozone-wayland recipes have been removes:
* Stop setting a few redundant variables.
* Drop the `gettext` class.
* Install some files unconditionally, remove a few subpackages that were only relevant for the CEF recipe.
* Stop generating empty an -dev package.
* Move all patches to `files/`.

See also: #62.


rakuco:
Note that the `README` is still in `recipes-browser/chromium`. It feels weird to put so much Chromium-specific text in the top-level `README`, but I can do that or move the text to the wiki if preferred.

otavio:
@rakuco I agree that adding the README on top-level README is awkward but please, in a next commit, at least put a reference on the top-level README about it so people know about its existence.

rakuco:
> but please, in a next commit, at least put a reference on the top-level README about it so people know about its existence.

Done in #68.

PR #77, opened 2017-11-28
chromium: unify recipe and move out upstream Chromium specific logic

msisov:
The intent of this change is to share as many parts of the recipe as
possible in order to easily switch between different upstream
and downstream versions of Chromium in the future.

To be more specific: chromium-gn.inc and chromium-x11.inc
are introduced.

chromium-gn.inc is used now for all the possible chromium
versions, and chromium-x11.inc will be shared between different
upstream Chromium versions. One just simply makes another
chromium-x11_{VERSION}.bb recipe and everything will be handled
automatically without dramatic changes.


recipes-browser/chromium/gn-native_62.0.3202.94.bb
```diff
@@ -14,6 +14,10 @@ S = "${WORKDIR}/chromium-${PV}"
 # evils.
 B = "${S}/out_bootstrap"
 
+
+SRC_URI[md5sum] = "98cf7b6eca255e2422f96094eccc1887"
+SRC_URI[sha256sum] = "cabc4d267bf08aabe11c5739048c43dde18c61acf595223a1c3aa1d3499558d4"
+
```
otavio: What about moving the checksum to chromium-upstream-tarball-${PV}.inc ? so we share this among multiple files and make it eaiser to upgrade.

recipes-browser/chromium/chromium-upstream-tarball.inc
```diff
@@ -0,0 +1,190 @@
...
+    file://${S}/third_party/wtl/Ms-PL.txt;md5=c1ff18e69b2e953f65f28bcbc29be275 \
+    file://${S}/third_party/xdg-utils/LICENSE;md5=a5367a90934098d6b05af3b746405014 \
+    file://${S}/third_party/yara/src/COPYING;md5=541962f9dacf27c928f57e3a7ba9e1f2 \
+    file://${S}/third_party/yasm/source/patched-yasm/COPYING;md5=a12d8903508fb6bfd49d8d82c6170dd9 \
+    file://${S}/third_party/zlib/LICENSE;md5=f09575dbfb09420642318b413159496f \
+    file://${S}/url/third_party/mozilla/LICENSE.txt;md5=437ced1e9b232651b0912a9594da43b2 \
+    file://${S}/v8/LICENSE;md5=71f2c485344c921f659ea9f7d895ea22 \
+    file://${S}/v8/LICENSE.fdlibm;md5=b17367d6c97e638599071d99a3049dfe \
+    file://${S}/v8/LICENSE.strongtalk;md5=956a43818dd13f2e93088fc2c93589d2 \
+    "
```
otavio: The checksum belongs here and it should be version specific.

recipes-browser/chromium/chromium-x11_62.0.3202.94.bb
```diff
@@ -0,0 +1,47 @@
...
+        file://musl-support/0017-Use-_fpstate-instead-of-_libc_fpstate.patch \
+        file://musl-support/0018-tcmalloc-Use-off64_t-insread-of-__off64_t.patch \
+        file://musl-support/0019-Disable-HAVE_SECURE_GETENV.patch \
+        file://musl-support/0020-Disable-mallinfo-usage-in-base.patch \
+"
+
+# Starting with M61, Chromium defaults to building with its own copy of libc++
+# instead of the system's libstdc++. Explicitly disable this behavior.
+# https://groups.google.com/a/chromium.org/d/msg/chromium-packagers/8aYO3me2SCE/SZ8pJXhZAwAJ
+GN_ARGS += "use_custom_libcxx=false"
```
otavio: Shouldn't this be common to all Chromium recipes? if so, it should be moved to chromium.inc

recipes-browser/chromium/chromium-upstream-tarball.inc
msisov: The idea is to share most of the checksums in one file, once there are upgrades and a previous version is kept for some reason, only changed checksums will be removed from here and moved to chromium-x11_{$PV}.bb. Thus, we can have as many upstream x11 versions as possible

recipes-browser/chromium/chromium-x11_62.0.3202.94.bb
msisov: Otherwise chromium-ozone-wayland, which is m64 fails to compile both in raspberry yocto and AGL environments.

recipes-browser/chromium/gn-native_62.0.3202.94.bb
msisov: The same gn version is used with m64 without any problems currently and the chromium-ozone-wayland recipe that we are going to upstream is using it without any problems. Wouldn't we end up with many chromium-upstream-tarball-${PV}.inc if we have different versions of chromium supported here? Those are chromium-x11 m62 and chromium ozone wayland m64. In any case, chromium-ozone-wayland m64 will be fetched from Igalia server with patches already applied. It is based on github.com/igalia/chromium ozone-wayland-dev branch

recipes-browser/chromium/chromium-upstream-tarball.inc
otavio: Sure. What I mean to move here are SRC_URI checksums. It can be shared and could be placed on a chromium-tarball-{version}.inc so it is included in the version-specific recipes using ${PV}.

recipes-browser/chromium/chromium-x11_62.0.3202.94.bb
otavio: I see. So m64 does not support this option?

recipes-browser/chromium/gn-native_62.0.3202.94.bb
otavio: I see your point. 

It is fine with me then. Let's wait few days and see if someone else comment.

msisov: ok!

recipes-browser/chromium/chromium-x11_62.0.3202.94.bb
msisov: there is such an option, but it requires patching the browser. 
And we haven't used this option for AGL builds and our internal yocto builds for rpi2 and rpi3. That's why we think of moving this out.

otavio: So it seems it should be used; this reduces the build time and also easy fixing of security issues if libc++ is affected. In the worse case, this should be moved to a PACKAGECONFIG.

msisov: Ok. I will work on this then and get to you back.

otavio: Just to reinforce the reasoning, it seems Debian also use this:

https://anonscm.debian.org/cgit/pkg-chromium/pkg-chromium.git/tree/debian/rules#n43

So keeping it enabled by default makes sense.

msisov: done!

kraj: disabling default C++ runtime for chromium is exception than norm. This also will enforce libstdc++ on to systems where its not expected e.g. if someone was using meta-clang and did not want to use libstdc++, so I like the idea of turning it into a packageconfig. If not then
lets keep using the chromium defaults for best results.

msisov: Will do so in the follow ups then

twoerner:
A while back (July 2016, in the chromium 48 days) when I split out the x11 and wayland parts I did the same thing, i.e. I created a patch where there was no "chromium" recipe but a "chromium-x11" and a "chromim-wayland" recipe. At that time it was felt that it would be better to have a "chromium" and a "chromium-wayland" recipe with the understanding that the default "chromium" recipe was for x11.

Is that no longer the case?
@kraj 

http://lists.openembedded.org/pipermail/openembedded-devel/2016-July/108289.html
http://lists.openembedded.org/pipermail/openembedded-devel/2016-July/108290.html

otavio:
Honestly, I hope that when Wayland work is merged upstream we can move back to a single recipe and just configure the backend we wish to use using PACKAGECONFIG. Until this happen user need to make a conscious decision about what to use.

I don't see a problem in adding the two recipes, for now. We can add a transition glue for current users, making chromium-x11 provide chromium so it is kept using the x11 as default. What do you think?

kraj:
I think it’s fine if rename happens but I think it’s better to keep the default be chromium whatever that is at a given time

msisov:
I better like the approach, when chromium is named with x11 prefix explicitly to avoid confusions.

msisov:
Chromium with Wayland support is being upstreamed at the moment and it will take some time. We will be updating the recipe accordingly

otavio:
@msisov I understand and currently having both possible recipes it does make sense. That said, we have people out there using the former name and keeping an upgrade path for them makes sense and do not adds complexity.

PR #79, opened 2017-11-30
Add ozone wayland recipe.

msisov:
This pull request adds chromium-ozone-wayland recipe, which is based on
https://github.com/Igalia/chromium. 

The Chromium browser in this recipe is based on the ToT Chromium and utilizes new ui service called mus with ozone x11 and wayland backends. The current support is being done by Igalia S.L., and is being upstreamed at the moment. In the future, all the work will be moved to upstream, which will simplify the recipe.

The recipe can be updated on a weekly or on a monthly basis depending on the status of the work.
Please note that the Chromium/Wayland can be affected by possible bugs in ToT and some of the features might be still missing. We are working to get Chromium/Wayland to be as similar to stock Chromium as possible. The work can be tracked at https://github.com/Igalia/chromium and crbug.com/666958


otavio:
@msisov I have a mix of feelings related to the nodejs patch. 

There is a typo in commitlog of chromium addition: introduce ;-)

@kraj what your view about adding nodejs here?

msisov:
@otavio, you're right. It's AGL specific. Let me check yocto poky for raspberry. If there is no such a problem, I'd better remove this patch from here and maintain it separately.

msisov:
UPD: yocto poky ships nodejs 8.9, which is ok for chromium. AGL ships 4.8.2, which is problematic. I'm removing the patch from this pr then.

msisov:
Can we do it other way? Let's have this master branch without the nodejs patch, but create a pyro branch, which will have nodejs update? Will it be ok then?

otavio:
The biggest problem in branching it now is that I believe this will be a moving target so the pyro branch would need to be constantly updated, which is not good for a stable branch.

I think AGL should do the backport in their layer (in a recipes-backport or similar) to allow them to use this new release while keep using an old version of OE-Core.

msisov:
ok

Issue #81, opened 2017-12-21
chromium-x11 crashes for use-egl at least


schnitzeltony:
Have tested chromium-x11 with quite recent master layers. It was build with

PACKAGECONFIG = "use-egl impl-side-painting proprietary-codecs"

for Raspi3.

On a fresh image chromium starts but as soon as I open e.g https://github.com it crashes. All further starts crash immediately. If I delete $HOME/.config/chromium I can start chromium again but moving to https://github.com crashes.

So I started chromium by console and get:

> [2263:2263:1221/015335.950665:ERROR:gl_implementation.cc(241)] Failed to load /usr/lib/chromium/libGLESv2.so: /usr/lib/chromium/libGLESv2.so: cannot open shared object file: No such file or directory
> [2263:2263:1221/015335.957888:ERROR:gpu_child_thread.cc(254)] Exiting GPU process due to errors during initialization
> [2213:2247:1221/015335.971247:ERROR:browser_gpu_channel_host_factory.cc(108)] Failed to launch GPU process.
> Bus error (core dumped)

I did some investigations:
* On Target: Creating a link /usr/lib/chromium/libGLESv2.so to /usr/lib//libGLESv2.so.2 did not change anything
* In OE-Workdir: Have no idea if this is something of importance but I found many entries like 'solibs = ./libEGL.so ./libGLESv2.so' in the created ninja files. The error message path is libexec / /usr/lib/chromium so this might cause chromium look there.

twoerner:
When you build chromium-x11 for the rpi3, when done you'll find the two files libGLESv2.so and libEGL.so left over in the build directory. If you install them to ${D}${libdir}/chromium you'll find that chromium now uses gpu acceleration when run. Not sure why this isn't the default. I.e.

do_install_append() {
    install libGLESv2.so ${D}${libdir}/chromium
    install libEGL.so ${D}${libdir}/chromium
}

I don't think I'm using anything other than the default PACKAGECONFIGs when I build.

msisov:
I guess PACKAGECONFIG must be looking the following
PACKAGECONFIG_pn-chromium-x11 = "use-egl impl-side-painting proprietary-codecs"

and then it will work. At least, I always specify.

schnitzeltony:
@msisov / @twoerner: Thanks for quick response

@msisov: I do this in an chromium%.bbappend file in my pseudo distro layer meta-misc. I think there the syntax should be ok.

@twoerner: I will give this a try - building will take some time...
[...]

PR #82, opened 2017-12-21
chromium: Install libEGL / libGLESv2 if created

schnitzeltony:
Fixes:
| [2263:2263:1221/015335.950665:ERROR:gl_implementation.cc(241)] Failed to load /usr/lib/chromium/libGLESv2.so: /usr/lib/chromium/libGLESv2.so: cannot open shared object file: No such file or directory
| [2263:2263:1221/015335.957888:ERROR:gpu_child_thread.cc(254)] Exiting GPU process due to errors during initialization
| [2213:2247:1221/015335.971247:ERROR:browser_gpu_channel_host_factory.cc(108)] Failed to launch GPU process.
| Bus error (core dumped)

Thanks to Trevor Woerner for suggesting this [1]

[1] https://github.com/OSSystems/meta-browser/issues/81

### DIFF
--- a/recipes-browser/chromium/chromium-gn.inc
+++ b/recipes-browser/chromium/chromium-gn.inc
@@ -384,6 +384,14 @@ do_install() {
 
 	# ChromeDriver.
 	install -m 0755 chromedriver ${D}${bindir}/chromedriver
+
+	# EGL / GLESv2
+	if [ -f libEGL.so ]; then
+		install libEGL.so ${D}${libdir}/chromium
+	fi
+	if [ -f libGLESv2.so ]; then
+		install libGLESv2.so ${D}${libdir}/chromium
+    fi
 }
 
 PACKAGES =+ "${PN}-chromedriver"


rakuco:
Chromium's scripts to generate the official Google Chrome packages on Linux have the following excerpt: https://chromium.googlesource.com/chromium/src.git/+/62.0.3202.94/chrome/installer/linux/common/installer.include#166

Which means those files aren't shipped in their stable packages, and, according to https://bugs.chromium.org/p/chromium/issues/detail?id=522133#c11, those files should end up being used only if chromium is started with `--use-gl=angle`.

So I'm a little skeptical of the approach described here. It'd be good to investigate whether ANGLE's being tried as a default GL backend, or why the other ones have not worked as expected, at least.

schnitzeltony:
Yeah I am not happy about it too - and it does not make chromium stable for me...

I checked:
<..>/chromium-x11/62.0.3202.94-r0/image/usr/lib/chromium/chromium-wrapper contains:

>CHROME_EXTRA_ARGS="         --use-gl=egl         --enable-gpu-rasterization --enable-impl-side-painting          "

Unfortunately up to now we have @schnitzeltony (me) and @twoerner facing this problem and both build for same machine - so it might be caused by Raspi3 / 32bit (@twoerner : which meta-raspberrypi do you use?) but I don't think so: GL is based on very recent mesa and images I use to build are close to world with lots of GLES/EGL used in X11 and wayland - but who knows..
Are there others seeing same problem for other machines?

Just a guess: Is it possible that other parts of build system consider build as not stable or build for angle (whatever that is :). Why is libEGL and libGLES build at all?

twoerner:
@rakuco I'm curious to know how this is this related to swiftshader?

@rakuco that's a good point regarding the --use-gl= flag. I had noticed that using --use-gl=egl caused GL acceleration to _not_ work, but removing the flag altogether makes it work.

twoerner:
@schnitzeltony I'm using the "default" one from @agherzan. 

Oh! and I build with gcc6. but gcc6 doesn't work with the libc from master, so I'm using rocko branches of openembedded-core and meta-openembedded. I seem to vaguely think that using gcc6 was somehow related to getting GPU acceleration working with chromium (but with my memory the way it is, maybe I'm mis-remembering that?).

I haven't tested with web pages from around the internet, we just use chromium to render some node stuff we have running locally on the rpi3-32. But that which we are rendering is working, and appears to be quite stable (as described above, i.e. gcc6 + rocko). We've run the device for weeks at a time without hickups.

rakuco:
> Just a guess: Is it possible that other parts of build system consider build as not stable or build for angle (whatever that is :). Why is libEGL and libGLES build at all?

I haven't taken a closer look at this part of the build system, but I think ANGLE is always built, but it's only shipped with the official packages for beta and dev releases. Not sure about upstream's plans regarding ANGLE on Linux though.

> @rakuco I'm curious to know how this is this related to swiftshader?

Also not sure if there's any relationship at all. AFAICS SwiftShader is x86-specific (at least it's only built on x86) and should be used when there's no hardware acceleration available, whereas ANGLE translates OpenGL ES calls to a multitude of other APIs (Desktop GL, DirectX etc).

> @rakuco that's a good point regarding the --use-gl= flag. I had noticed that using --use-gl=egl caused GL acceleration to not work, but removing the flag altogether makes it work.

Does not passing `--use-gl=egl` make things Just Work (TM) for you? How about @msisov's comment in #81 regarding vc4graphics?

twoerner:
@rakuco hmm... all good questions. I'm going to do some builds and runtime testing before getting back.

twoerner:
Here's what my testing has revealed to me.

Build Configuration:

```
> BB_VERSION           = "1.37.0"
> BUILD_SYS            = "x86_64-linux"
> NATIVELSBSTRING      = "opensuse-42.3"
> TARGET_SYS           = "arm-oe-linux-gnueabi"
> MACHINE              = "raspberrypi3"
> DISTRO               = "nodistro"
> DISTRO_VERSION       = "nodistro.0"
> TUNE_FEATURES        = "arm armv7ve vfp thumb neon vfpv4 callconvention-hard cortexa7"
> TARGET_FPU           = "hard"
> meta-raspberrypi     = "master:24ffe9fb592dda483fb5dfec30782955fcc18a5e"
> meta                 = "master:d214b1dc9339c4873156934f5291b9373a11471f"
> meta-oe              = "master:acfb8417e370d017e7307872e61564deeda76a13"
> meta-browser         = "master:4dda3fb94072a515763d816e15855d7fdf210a25"
```


Building off master, with all the default PACKAGECONFIGs, with gcc7, ...

If I don't provide "--with-gl=egl" (and delete it from the wrapper script) then I get nothing on the screen and the message:

> `Modifier 0x0 vs. tiling (0x700000000000001) mismatch`

If I provide "--with-gl=egl" then I get the app come up on the screen, but GPU is not accelerated, and the message:

```
> [1124:1124:1221/191951.692029:ERROR:gl_implementation.cc(241)] Failed to load /usr/lib/chromium/libGLESv2.so: /usr/lib/chromium/libGLESv2.so: cannot open s
> hared object file: No such file or directory
> [1124:1124:1221/191951.700093:ERROR:gpu_child_thread.cc(254)] Exiting GPU process due to errors during initialization
> [1089:1116:1221/191951.718258:ERROR:browser_gpu_channel_host_factory.cc(108)] Failed to launch GPU process.
```

If I copy the system libEGL.so and libGLESv2.so to /usr/lib/chromium then I get nothing on the screen and the dreaded:

> `Modifier 0x0 vs. tiling (0x700000000000001) mismatch`

If I copy the libEGL.so and libGLESv2.so from the chromium build to /usr/lib/chromium I also get nothing on the screen and the dreaded:

> `Modifier 0x0 vs. tiling (0x700000000000001) mismatch`


If I build with gcc6...

First off, the master branch no longer provides gcc6. This is a change from a couple weeks ago; at that point gcc6 was provided, but libc failed to ./configure. I guess someone noticed gcc6 wasn't working and removed it altogether. This means I had to switch to the latest release branches (which at this point in time means rocko).

Build configuration:

```
> BB_VERSION           = "1.37.0"
> BUILD_SYS            = "x86_64-linux"
> NATIVELSBSTRING      = "opensuse-42.3"
> TARGET_SYS           = "arm-oe-linux-gnueabi"
> MACHINE              = "raspberrypi3"
> DISTRO               = "nodistro"
> DISTRO_VERSION       = "nodistro.0"
> TUNE_FEATURES        = "arm armv7ve vfp thumb neon vfpv4 callconvention-hard cortexa7"
> TARGET_FPU           = "hard"
> meta-raspberrypi     = "master:24ffe9fb592dda483fb5dfec30782955fcc18a5e"
> meta                 = "rocko:39fc1150e74173fc6d73c464c141dba0b7ac2930"
> meta-oe              
> meta-python          = "rocko:6e3fc5b8d904d06e3aa77e9ec9968ab37a798188"
> meta-browser         = "master:4dda3fb94072a515763d816e15855d7fdf210a25"
```

If I run chromium with no cmdline arguments (which includes removing all cmdline arguments from the wrapper script) chromium starts and the GPU is accelerated (as per chrome://gpu).

If I provide --use-gl=egl chromium starts but complains about missing EGL/GLESv2 and is therefore not accelerated:
```
> [1094:1094:1221/223040.290602:ERROR:gl_implementation.cc(241)] Failed to load /usr/lib/chromium/libGLESv2.so: /usr/lib/chromium/libGLESv2.so: cannot open s
> hared object file: No such file or directory
> [1094:1094:1221/223040.303149:ERROR:gpu_child_thread.cc(254)] Exiting GPU process due to errors during initialization
> [1057:1084:1221/223040.321904:ERROR:browser_gpu_channel_host_factory.cc(108)] Failed to launch GPU process.
```

If I copy the EGL/GLES libraries that are built as part of chromium and copy them to the target's /usr/lib/chromium then I can start chromium with --use-gl=egl, there are no complaints, and the rendering is GPU accelerated.


IN CONCLUSION:
- building with gcc7: the binary fails to display anything on the screen (modifier vs tiling error) unless I specify --use-gl=egl but don't copy any EGL/GLES binaries to /usr/lib/chromium
- building with gcc6 (requires downgrading to rocko branch): the binary always displays on screen and will either be accelerated or not; oddly enough, without copying any EGL/GLES libraries to /usr/lib/chromium and without specifying any cmdline arguments, chromium appears to be GPU accelerated

rakuco:
Thanks for the thorough investigation, Trevor. I only have one question left: what if you build Chromium on rocko with GCC 7?
@msisov @jdapena does any of this sound familiar to you in your RPi3 tests?

otavio:
It is not yet clear if this is the right approach. Close this as there is a open bug related.

PR #83, opened 2017-12-21
chromium-x11: Set PROVIDES="chromium".

rakuco:
As discussed in pull request #77, it makes sense to provide some
compatibility glue for users who were used to just doing "bitbake chromium"
before the recipe got renamed to "chromium-x11".


rakuco:
+@msisov

PR #84, opened 2017-12-21
chromium: Post-ozone-wayland cleanups

rakuco:
This PR includes several clean-ups after the ozone-wayland-related pull requests were merged a few weeks ago.

They mainly reduce duplication and drop a few `.inc` files to reduce confusion.


rakuco:
@otavio @msisov

otavio:
The jumbo build could be configured somehow, don't you think? If so, what we default?

rakuco:
> The jumbo build could be configured somehow, don't you think? If so, what we default?

It can; one just needs to append `use_jumbo_build=true` to `GN_ARGS` locally. By default, I'd say we follow upstream (which sets it to false).

recipes-browser/chromium/chromium-ozone-wayland_64.0.3274.0.r517731.igalia.1.bb
```diff
@@ -1,8 +1,25 @@
...
+        libxkbcommon \
+        virtual/egl \
+        wayland \
+"
+
+GN_ARGS += "\
+        enable_mus=true \
+        use_ozone=true \
+        ozone_auto_platforms=false \
+        ozone_platform_headless=true \
```
msisov: can you set it false as well? I believe it's not needed.

rakuco: Would you be OK with doing that in a separate PR? Changing the value of the option causes the build to change, whereas I was hoping the cleanups here would have no effect on the actual build.

msisov: please go ahead as you wish!

rakuco: ... though arguably removing `use_jumbo_build=true` does change how things are built... I'll send that part in a separate PR too.

rakuco:
@msisov @otavio I've pushed a new version that doesn't change the Jumbo builds part. There shouldn't be anything controversial now.

PR #85, opened 2017-12-22
chromium-ozone-wayland: Disable Jumbo builds by default.

rakuco:
Whether to use Jumbo builds or not should be up to each user; while it may
decrease build times, it can require more RAM (it tends to cause problems
for me specifically when using GCC because of the amount of memory it
requires).

Additionally, it's not an Ozone-Wayland-specific setting, so it should've
been in chromium-gn.inc had it been the case.

We now provide a PACKAGECONFIG knob (off by default) to control that.

### DIFF
--- a/recipes-browser/chromium/README
+++ b/recipes-browser/chromium/README
@@ -45,6 +45,14 @@ PACKAGECONFIG knobs
   http://www.chromium.org/developers/design-documents/impl-side-painting for
   more.
 
+* jumbo-build: (off by default)
+  Enables what's also known as "unity builds": several source files are merged
+  together and compiled only once in order to reduce build times, but may end
+  up requiring more memory at certain points.
+
+  Jumbo builds are documented here:
+  https://chromium.googlesource.com/chromium/src/+/master/docs/jumbo.md
+
 * kiosk-mode: (off by default)
   Enable this option if you want your browser to start up full-screen, without
   any menu bars, without any clutter, and without any initial start-up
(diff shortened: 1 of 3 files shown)


rakuco:
@msisov @otavio

schnitzeltony:
Maybe I am wrong here: You write changes are applied in chromium-gn.inc but the patch applies on chromium-ozone-wayland only. What did I miss?

rakuco:
I write that, should we decide to pass `use_jumbo_build=true` to GN, it should have been done in chromium-gn.inc, not the oz-wl .bb recipe file.

schnitzeltony:
Ahh - not enough coffee...

If you make it common, I promise to play around with it for ..x11 :)

rakuco:
Well, the idea is not to touch the setting at all, but allow anyone who'd like to enable it to do so in their own `local.conf`. That applies to the x11 version too.

recipes-browser/chromium/README
```diff
@@ -26,6 +26,22 @@ version 5.4.0.
...
+
+Jumbo builds are documented here:
+https://chromium.googlesource.com/chromium/src/+/master/docs/jumbo.md
+
+We follow Chromium's default settings, and also disable Jumbo builds by default.
+
+If you would like to enable Jumbo builds locally, you can pass additional flags
+to GN (in local.conf, for example):
+
+    GN_ARGS_append_pn-chromium-x11 = " use_jumbo_build=true"
```
otavio: I understand that this is what happens behind the scene but I think we ought to use the `PACKAGECONFIG` mechanism to change this as this is the standard way of configuring the package.

Doing so, adding a comment on the recipe before the block adding the config does help.

rakuco: I thought `PACKAGECONFIG` was normally used for knobs which have an effect on the generated package (support for additional features, changes to the build or runtime dependencies).

Jumbo builds only affect the build itself, not what is generated in it.

I'm fine with `PACKAGECONFIG`, but I wanted to double-check we're not abusing it.

otavio: It is intended to configure the package. It can be for:

* enabling / disabling a feature
* change behaviour

In this case, it enables the jumbo build support. Fits well IMO.

rakuco: Fair enough; I've switched to a `PACKAGECONFIG` knob, please take another look.

PR #88, opened 2017-12-22
chromium-x11: Update to 63.0.3239.108.

rakuco:
This is the latest stable release at the time of writing.

From a recipe point of view, there are no big changes: several patches are
no longer needed and a few backports were added.

0001-Use-v8-qemu-wrapper.sh-on-v8-context-snapshot-creati.patch, which used
to be specific to the chromium-ozone-wayland recipe, has now been merged
into the main v8-qemu-wrapper.patch file, as both M63 and M64 build Blink's
v8_context_snapshot.bin.

Tested with:
* pyro, GCC 5.4 and MACHINE="raspberrypi3"
* rocko, GCC 7.2.1, MACHINE="qemux86-64" and MACHINE="qemuarm64"


rakuco:
@twoerner @otavio @msisov 

I haven't tried building chromium-ozone-wayland, so some feedback there would be appreciated, as well as additional testing + running on multiple compiler/platform configurations.

otavio:
@rakuco I am fine with all changes; I think it is just important to have a check from Wayland recipe that it at least continue building.

Is it possible for you to check it? @msisov it would be great if you guys could review this as well (and possibly refresh the wayland recipe too?)

otavio:
@twoerner @schnitzeltony possibly this helps with #82 ?

rakuco:
Not sure about #82, but I'll create another build directory for chromium-ozone-wayland and come back to this after the holidays (next Wednesday).

schnitzeltony:
I would close #82. It does not solve the problem and it unclear if we really need it. Of course I'll try to help out solving but first I need to get remote debugging tested for recipe specific sysroot then I need to find a way to get debug symbols into chromium again. From my experience debugging is the only way to get helpful information why chromium crashes. For further discussion I suggest #81.

OK?

rakuco:
Works for me, thanks for helping with the investigation!

rakuco:
> but I'll create another build directory for chromium-ozone-wayland and come back to this after the holidays (next Wednesday).

Alright, I've built `chromium-ozone-wayland` on rocko with `MACHINE="beaglebone"` and everything went fine.

PR #92, opened 2018-01-25
chromium-ozone-wayland: Replace the local memcpy patch with the upstream one

rakuco:
It turns out `0001-Fix-memcpy-was-not-declared-in-this-scope.patch` was never
sent upstream, but ended up being fixed independently last month.

Replace the local version with the upstream commit, which is pretty much
identical.


rakuco:
@msisov @otavio

PR #93, opened 2018-01-26
chromium-x11: Update to 64.0.3282.119.

rakuco:
Release notes:
https://chromereleases.googleblog.com/2018/01/stable-channel-update-for-desktop_24.html

Drop a few backported patches that are part of this release, and backport a
few new ones to fix the build with GCC 5 and with the CUPS option disabled.

The chromium-x11 recipe is now newer than the chromium-ozone-wayland recipe,
and v8-qemu-wrapper.patch has diverged between the two versions. The old
version had to be temporarily copied to a separate patch directory so that
the ozone-wayland recipe continues to build. There are plans to update it
soon, so the tree shouldn't look too ugly for too long.

Test-built in:
- pyro with GCC 5 and MACHINE="raspberrypi3"
- rocko with GCC 7 and MACHINE="qemux86-64", MACHINE="qemuarm64" and
  MACHINE="beaglebone" (the latter only for the chromium-ozone-wayland
  recipe).


rakuco:
@otavio @msisov

msisov:
lgtm. I'm going to update the wayland recipe soon as well.

otavio:
Nice work :-)

PR #95, opened 2018-02-02
chromium-x11: Update to 64.0.3282.140.

rakuco:
Release notes:
https://chromereleases.googleblog.com/2018/02/stable-channel-update-for-desktop.html

Test-built on:
* pyro with `MACHINE="raspberrypi3"` and GCC 5
* rocko with `MACHINE="qemux86-64"` with GCC 7.

chromium-ozone-wayland has also been verified to continue working with GCC 7
and `MACHINE="beaglebone"`.


kraj:
why do you stick to gcc5 for rpi3 ? thats a concern

rakuco:
> why do you stick to gcc5 for rpi3 ? thats a concern

It's not specifically for the rpi3 build, but rather that I wanted to continue testing GCC 5 for _some_ ARM configuration. At least in the past it was important to remain compatible with GCC 5 for Igalia to merge their own meta-browser recipe into this one (cc @tonikitoo @msisov), and keeping that working hasn't been much of a hassle.

Plus, if something works with both GCC 5 and GCC 7 there's also a pretty high chance that it will also work with GCC 6.

PR #106, opened 2018-03-21
chromium-x11: Remove musl-related code.

rakuco:
musl support has not been tested in a very long time, and the patches very
likely no longer apply. It only makes sense to keep musl patches around if
there is work to upstream them, otherwise we end up with an increasing
amount of patches that touch sensitive, low-level areas of the code to
maintain.


otavio:
@kraj could you check if musl is still feasible?

PR #107, opened 2018-04-06
Patch refresh / add LAYERSERIES_COMPAT

schnitzeltony:


rakuco:
Was the Chromium patch failing to apply to you before? I can see that the context looked wrong indeed (so thanks for the fix), but I didn't have any trouble applying it before.

schnitzeltony:
It was applying but gave the 'fuzz' warning which is created with recent oe-core. I should have mentioned that in commit message - sorry.

PR #113, opened 2018-05-07
chromium-x11: Remove 0001-Fix-build-with-glibc-2.27.patch

rakuco:
This patch is already part of 66.0.3359.139, so we were just doing a no-op
when trying to apply it.


rakuco:
Hmm, this is true for at least one other patch. I'll send another PR covering everything.

PR #114, opened 2018-05-07
chromium-x11: Remove patches that are already part of the source

rakuco:
Drop a few patches were already part of 66.0.3359.139 but still included in
the recipe (as there were no complaints from `do_patch` about them):
* 0001-Add-missing-stdint-include.patch
* 0001-Fix-build-with-glibc-2.27.patch


twoerner:
Excellent! Have you tried building against master? Recently some changes were made to master such that if a patch applies with fuzz, OE will complain about it (but continue to work). Many of the chromium-x11 patches currently apply with fuzz. Removing patches that are no longer needed is great, but have you noticed any of the "patch applied with fuzz" warnings too?

rakuco:
I haven't tried building against master yet; I thought @schnitzeltony had fixed the warnings in #107, but apparently there are others that need to be dealt with.

While at it: is there a sane way to refresh the patches? I usually `git cherry-pick` them directly from upstream, which explains the fuzz warnings, and I hope fixing those warnings doesn't mean having to manually apply them and re-create the diffs manually :-)

twoerner:
I'm doing a fresh build with master now. The number of fuzz warnings has certainly gone down. The current fuzz warnings are now:

```
Applying patch 0001-GCC-PlaybackImageProvider-Settings-explicitely-set-c.patch
patching file cc/raster/playback_image_provider.cc
patching file cc/raster/playback_image_provider.h
Hunk #2 succeeded at 36 with fuzz 1 (offset -5 lines).

Applying patch 0001-GCC-build-fix-base-Optional-T-requires-the-full-decl.patch
patching file services/preferences/tracked/pref_hash_filter.h
Hunk #1 succeeded at 21 with fuzz 1.

Applying patch 0001-GCC-fully-declare-ConfigurationPolicyProvider.patch
patching file components/policy/core/browser/browser_policy_connector_base.h
Hunk #1 succeeded at 11 with fuzz 2 (offset -1 lines).
```

According to the text output by OE of the fuzz issue:

```
The context lines in the patches can be updated with devtool:

    devtool modify <recipe>
    devtool finish --force-patch-refresh <recipe> <layer_path>
```

but the last time I tried that, it failed. I'll give it another try when this build is complete and let you know how it goes.

schnitzeltony:
> I haven't tried building against master yet; I thought @schnitzeltony had fixed the warnings in #107, but apparently there are others that need to be dealt with.

FYI: At the time of #107 all warnings were gone for me. Since then there were some upgrades which introduced new warnings. Wonder what the patches change when already applied upstream...

twoerner:
I was able to perform the

```
devtool modify <recipe>
devtool finish --force-patch-refresh <recipe> <layer_path>
```

and the patches have been updated. However, in its infinite wisdom :-), devtool has stripped off the original email headings. In other words, the part of the patch that describes the original patch author

```
From: Jose Dapena Paz <jose.dapena@lge.com>
...
```

has been replaced with

```
From: Raphael Kubo da Costa <raphael.kubo.da.costa@intel.com>
...
```

I'll restore the original authors, do a test with the new patches, and submit.

otavio:
@twoerner great! :-)

twoerner:
Please see https://github.com/OSSystems/meta-browser/pull/115 for the patch updates. With that patch, building off master no longer generates any patch fuzz warnings.

However, I do understand what @rakuco is saying, hopefully "fixing" the patches doesn't mean he now has to tweak them every time he grabs them from upstream. If that is the case, my patch to "fix" the patches shouldn't be applied. I'd rather see fuzz warnings rather than increase the maintenance burden.

twoerner:
I would only run those two devtool commands if bitbake told me to (i.e. if it detected patches being applied with fuzz). If bitbake doesn't produce any warnings, all is good :-)

PR #115, opened 2018-05-08
chromium-x11: patch refresh

twoerner:
Bitbake will now flag patches that require fuzz when applied:

	http://lists.openembedded.org/pipermail/openembedded-core/2018-March/148675.html
	https://bugzilla.yoctoproject.org/show_bug.cgi?id=10450

Fixup the affected patches.

Signed-off-by: Trevor Woerner <twoerner@gmail.com>


rakuco:
Can you just include some instructions on how to refresh those patches in the commit message? It's not required, but it helps me not mess things up when I update the recipe.

twoerner:
Okay, instructions now included.

rakuco:
lgtm, thanks!

PR #117, opened 2018-05-10
firefox: Remove musl support

cosmo0920:
PR #106 removes musl support.
But firefox recipe still support musl build.
Can we remove musl support from firefox?

We're planning to upgrade firefox recipe to 52esr like as [@ashie's work](https://github.com/OSSystems/meta-browser/pull/43).

This is one of the blockers to upgrade firefox into 52esr from 45esr.


cosmo0920:
@kraj Do you still need musl support on firefox?

cosmo0920:
Wow, thank you!

kraj:
We have been lagging on musl support, I think it needs to be redone and I would then expect that community also helps in keeping it updated, its not a blocker for now IMO

PR #120, opened 2018-05-23
firefox: Use --disable-elf-hack option if supported

cosmo0920:
--disable-elf-hack option should specify with EXTRA_OECONF overrides.
Always specifying it causes confusion why disable-elf-hack is not effective
such as aarch64 and mips and so on.

Because Elf-hack supports arm, i586, and x86_64.
Other platforms is not supported and ignored such as aarch64(Arm 64bit).

ref: https://dxr.mozilla.org/mozilla-esr45/source/configure.in#7398

Signed-off-by: Hiroshi Hatake <hatake@clear-code.com>

### DIFF
--- a/classes/mozilla.bbclass
+++ b/classes/mozilla.bbclass
@@ -7,7 +7,10 @@ inherit gettext pkgconfig
 
 EXTRA_OEMAKE += "SHELL=/bin/sh"
 EXTRA_OECONF = "--target=${TARGET_SYS} --host=${BUILD_SYS} \
-                --build=${BUILD_SYS} --prefix=${prefix} --disable-elf-hack"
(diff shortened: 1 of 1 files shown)


otavio:
@cosmo0920 there few minor things to improve.

Instead of using a function to detect the arch, use overrides as:

```
EXTRA_OECONF_arm_append = " --disable-elf-hack"
EXTRA_OECONF_x86_append = " --disable-elf-hack"
```

I am unsure if x86_64 is covered by the generic override, if it is not, add it as well.

The commit log also need some improvement. It should mention *why* adding the flag is important, besides the architecture restrictness. The last bit is the missing signed-off-by field.

PR #128, opened 2018-06-14
[Do not merge] Upgrade to Firefox 60ESR

ashie:
(We confirmed this reciepe only on Renesas RZ/G and R-Car.)

Although we've just contributed the recipe for Firefox 52ESR recently (#121), we already have the recipe for 60ESR. We also want to contribute it, there are some big issues to merge it. We need to discuss it.

The biggest issue is introducing Rust compiler.
60ESR requires Rust 1.23 or later.
There are some ways to introduce it.

* Use [meta-rust](https://github.com/meta-rust/meta-rust)
  * Need to modify to introduce crates (libraries) for target architectures (FYI: our PR which is rejected: https://github.com/meta-rust/meta-rust/pull/191)
* Use [meta-rust-bin](https://github.com/rust-embedded/meta-rust-bin)
  * Need to modify
* Use [rustup](https://rustup.rs/) as a host side tool

This pull request uses the 1st way to resolve it.
(We also confirmed that the 3rd way can build it successfully.)
We have our modified version of meta-rust to introduce std crate for our target architecture.

* For ARM: https://github.com/webdino/meta-rust/tree/jethro-14.0.1_rust_1.24.1
* For aarch64: https://github.com/webdino/meta-rust/tree/jethro-14.0.1_rust_1.24.1-aarch64

If we choose this way, we need to merge them to upstream before updating to 60ESR.
Even if we achieve it, meta-browser depends on meta-rust as a result.
Users of chromium won't accept it because it doesn't depend on it.
How should we resolve it?

I think the best way to resolve it is split them to each layers under meta-browser like this:

  * meta-browser
    * meta-chromium
      * recipes-chromium
    * meta-mozilla
      * recipes-mozilla

Or if we choose the 3rd way (use rustup), we don't need to split the layer.
Although the users that want to build firefox need to introduce rustup, the users that want to build
chromium doesn't need it.

How do you think about it?


ashie:
> * Use [meta-rust](https://github.com/meta-rust/meta-rust)
>   * Need to modify to introduce crates (libraries) for target architectures (FYI: our PR which is rejected: 

If your target is x86, you don't need to modify it.

ashie:
This recipe includes following upstream's backport patches:

  * Support EGL on Wayland
  * Improve Wayland support (such as DnD, Copy and Paste)
  * Support OpenMAX

Wayland support itself is already merged into 60ESR, so you don't need to additional patch to enable it.

otavio:
Hello @ashie 

First, I'd like to thank you for the work on this. This is a very welcome addition.

I don't foresee a problem in adding a dependency on meta-rust as, if unused, causes no problem for other people. Chromium users won't have its build result changed because of that so it should be fine.

One thing it is puzzling me is the PR in meta-rust. I did not understand what the problem is. So few things here:

- [ ] please rebase this PR on top of our master branch
- [ ] squash the commits in a logical patchset
- [ ] finish the PR work for meta-rust/meta-rust#191

ashie:
Thank you for you reply :)

> I don't foresee a problem in adding a dependency on meta-rust as, if unused, causes no problem for > other people. Chromium users won't have its build result changed because of that so it should be fine.

Just to confirm, Chromium users also need to add meta-rust because bitbake will check dependency
of all recipes on starting up. Is it O.K.?
If it's O.K. for you, I'll go forward this way.

> One thing it is puzzling me is the PR in meta-rust. I did not understand what the problem is. 

To build Rust code, Rust's std library (named "crate" in Rust's world) for a target architecture is required.
But current meta-rust supports only x86 architecture.
When a user build it for other architectures, it will fail.
We have to add other architecture support (at least ARM) to meta-rust.
Our previous PR against meta-rust cares only ARM, so it breaks x86 architecture support.
Probably we can resolve this issue by adding each architecture's settings like this:

* https://github.com/webdino/meta-rust/commit/9421cbcce21340a68e4d59f7ac39b8cfcc38d463

> So few things here:
>  * [ ] please rebase this PR on top of our master branch 
>  * [ ] squash the commits in a logical patchset
>  * [ ] finish the PR work for meta-rust/meta-rust#191

I see.
Of course we'll do it before we send a formal PR.
In addition to them, I think following tasks are also needed:

* [ ] Remove some evil patches
* [ ] Import Debian's patch set as usual

otavio:
> Just to confirm, Chromium users also need to add meta-rust because bitbake will check dependency
> of all recipes on starting up. Is it O.K.?

Yes. We need to live with that.

So please proceed fixing Rust support and reworking the PR :-)

rakuco:
If I understood it correctly, from a Chromium perspective one might need to add meta-rust to bblayers.conf, but there's nothing else that's changing right?

ashie:
> If I understood it correctly, from a Chromium perspective one might need to add meta-rust to bblayers.conf, but there's nothing else that's changing right?

That's right.

rakuco:
Then I definitely don't see why not :-)

ashie:
> So please update this PR when you go over the discussed pending items.

Can I open a new pull request with another branch for it?
I don't want force update this branch because this branch is already published to our users.

otavio:
Sure, I'm closing it then.

PR #132, opened 2018-07-10
chromium-x11: Update to 67.0.3396.99

rakuco:
Release announcement:
https://chromereleases.googleblog.com/2018/06/stable-channel-update-for-desktop_25.html

Test-built with:
* Yocto pyro, GCC 6 and MACHINE="raspberrypi3"
* Yocto rocko, GCC 7 and MACHINE="qemuarm64"
* Yocto sumo, GCC 7 and MACHINE="intel-corei7-64"

There are no changes to GN, so chromium-ozone-wayland should continue to be
fine.


rakuco:
Just wider coverage, I'm trying to cover as many architectures and compiler versions as possible.

kraj:
OK. Then you should change one of them to use gcc8 as well with  OE-Core master.

rakuco:
As time and disk space allows :-)

twoerner:
I think there are a couple of us who test rpi with chromium on master (i.e. gcc8).
Personally I have a jenkins job that does so once a week (rpi3-32 with vc4 on x11).
I haven't done a run-time test in a while, but I do run-time test it now and again.

PR #142, opened 2018-07-23
chromium-x11: Fix entangled cross/native build flags

kraj:
When we use pkgconf instead of pkg-config all sorts of
cflags issues start to appear resulting in compiler errors
which all the due to mixing of target cflags and ldflags
onto native compiler flags.

This patch tries to ensure that target flags are contained
to cross builds alone. This issue is also evident when building
for non-glibc targets e.g. where headers will be different
when compared to build host ( which mostly are glibc based)

### DIFF
--- a/recipes-browser/chromium/chromium-x11_67.0.3396.99.bb
+++ b/recipes-browser/chromium/chromium-x11_67.0.3396.99.bb
@@ -6,7 +6,13 @@ SRC_URI += "\
         file://0001-Ensure-all-targets-build-when-target_arch-arm-and-target_os-linux.patch \
         file://0001-Fix-operator-bool-in-AssociatedInterfacePtrInfo-and-.patch \
         file://aarch64-skia-build-fix.patch \
+        file://cross-build-pkgconfig.patch \
 "
+
+BUILD_CPPFLAGS += "-isystem${STAGING_INCDIR_NATIVE}/glib-2.0 \
+                   -isystem${STAGING_LIBDIR_NATIVE}/glib-2.0/include \
+                   -isystem${STAGING_INCDIR_NATIVE}/nss3 \
+                  "
 
 REQUIRED_DISTRO_FEATURES = "x11"
 
--- a/recipes-browser/chromium/files/cross-build-pkgconfig.patch
+++ b/recipes-browser/chromium/files/cross-build-pkgconfig.patch
@@ -1,0 +1,85 @@
+Chromium build system currently, mixes the paths for native and
+cross builds, where it detects glib-2.0 on target system using
+pkg-config but then tries to use the same for native portions of
+the build as well. In all it is expecting a non-empty sysroot for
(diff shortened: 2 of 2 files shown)


rakuco:
Thanks, I'll take a proper look at this tomorrow.

kraj:
No it does not help

rakuco:
It looked OK here; what steps do I need to follow to reproduce the pkgconf issue you mentioned in the patch description? I'd like to solve this with GN options if possible, maintaining that `BUILD_CPPFLAGS` list manually will not scale.

rakuco:
Thanks, can I add those local.conf bits to a sumo build directory or do I need to track master?

rakuco:
Right, using pkgconf on sumo is enough to reproduce the issue you describe.

As far as I can see, it's either a bug in pkgconf or in the way we handle its differences compared to pkg-config in OE: pkgconf should not spew `-I${STAGING_INCDIR}` as part of `--cflags` (ditto for the lib counterpart of this issue).

For example, `libpcre.pc` has `Cflags: -I${includedir}`. Since we set `PKG_CONFIG_SYSROOT_DIR=${RECIPE_SYSROOT}`, this becomes `-I${RECIPE_SYSROOT}/usr/include` aka `-I${STAGING_INCDIR}`. Invoking `pkg-config --cflags libpcre` from devshell produces an empty output with freedesktop's pkg-config and `-I${STAGING_INCDIR}` with pkgconf. That's because when deciding whether to filter out a certain `-I` parameter pkgconf is comparing `/usr/include` and `${RECIPE_SYSROOT}/usr/include` and failing to omit a system include directory. Solving this should solve the build issues you are seeing automagically as far as I can see. Another way to look at it: pkgconf seems to expect `PKG_CONFIG_SYSTEM_INCLUDE_PATH` to include the sysroot path, pkg-config does not. I think pkg-config is doing the right thing here, but I couldn't find a canonical source of truth for the expected behavior here.

A test build with pkgconf and `PKG_CONFIG_SYSTEM_{INCLUDE,LIBRARY}_PATH` with the sysroot prepended worked just fine here (though I had to stop requiring `security_flags.inc` due to errors like "cc1plus: error: -Wformat-security ignored without -Wformat [-Werror=format-security]" in the Chromium build).

(the `host_pkg_config` part should go in regardless of what's discussed here, I'll take care of it separately)

kraj:
lets close this one for now

rakuco:
Alright. If you're driving any work in OE to replace pkg-config with pkgconf by default, it'd be good to raise this issue with the pkgconf guys upstream.

kraj:
yes its sort of known to them now and there are defects open https://git.dereferenced.org/pkgconf/pkgconf/issues/2

rakuco:
I don't think it's the same issue. `PKG_CONFIG_SYSROOT_DIR` and others are already supported by pkgconf, it's just that the implementation looks buggy.

PR #146, opened 2018-07-25
chromium: Use pkg-config-native when building code on the host.

rakuco:
So far "pkg-config" was being used both for host and target toolchains; in
practice this meant we would end up passing directories belonging to the
target sysroot when compiling things for the host side.

This has not caused any actual build issues, but it is definitely wrong.


rakuco:
(originally discussed in #142)

PR #150, opened 2018-08-01
chromium-x11: Update to 68.0.3440.84.

rakuco:
Release announcement:
https://chromereleases.googleblog.com/2018/07/stable-channel-update-for-desktop_31.html

Test-built with:
* Yocto pyro, GCC 6 and MACHINE="raspberrypi3"
* Yocto rocko, GCC 7 and MACHINE="qemuarm64"
* Yocto sumo, GCC 7 and MACHINE="intel-corei7-64"

There are no changes to GN, so chromium-ozone-wayland should continue to be
fine.


kraj:
LGTM

Issue #135, opened 2018-07-16
chromium-x11: Consider dropping support for GCC 6 (and Yocto pyro)


rakuco:
I'm currently working on Chromium 68 (currently in the beta channel), and both ARM and ARM64 builds (I've tried `MACHINE="qemuarm64"` and `MACHINE="raspberrypi3"`) currently fail with pyro's GCC 6 due to an internal compiler error when building some bits of Skia that use NEON and other optimizations:

```
arch64-poky-linux-gcc  --sysroot=/data/src/yocto/poky/build-pyro-qemuarm64/tmp/work/aarch64-poky-linux/chromium-x11/68.0.3440.59-r0/recipe-sysroot -MMD -MF obj/skia/skcm
s/Transform.o.d -DV8_DEPRECATION_WARNINGS -DUSE_UDEV -DUSE_AURA=1 -DUSE_GLIB=1 -DUSE_NSS_CERTS=1 -DUSE_X11=1 -DNO_TCMALLOC -DFULL_SAFE_BROWSING -DSAFE_BROWSING_CSD -DSAFE
_BROWSING_DB_LOCAL -DOFFICIAL_BUILD -DCHROMIUM_BUILD -D_FILE_OFFSET_BITS=64 -D_LARGEFILE_SOURCE -D_LARGEFILE64_SOURCE -D__STDC_CONSTANT_MACROS -D__STDC_FORMAT_MACROS -D_F
ORTIFY_SOURCE=2 -DNDEBUG -DNVALGRIND -DDYNAMIC_ANNOTATIONS_ENABLED=0 -I../.. -Igen -I../../third_party/skia/third_party/skcms -w -std=c11 -fno-strict-aliasing --param=ssp
-buffer-size=4 -fstack-protector -funwind-tables -fPIC -pipe -pthread -Wall -Wno-unused-local-typedefs -Wno-maybe-uninitialized -Wno-deprecated-declarations -fno-delete-n
ull-pointer-checks -Wno-comments -Wno-missing-field-initializers -Wno-unused-parameter -O2 -fno-ident -fdata-sections -ffunction-sections -fno-omit-frame-pointer -g0 -fvi
sibility=hidden -std=gnu11  -O2 -pipe -feliminate-unused-debug-types -fdebug-prefix-map=/data/src/yocto/poky/build-pyro-qemuarm64/tmp/work/aarch64-poky-linux/chromium-x11
[...]

PR #103, opened 2018-03-09
chromium-x11: Update to 65.0.3325.146.

rakuco:
chromium-x11: Update to 65.0.3325.146.

This is the first M65 stable release. Annonuncement:
https://chromereleases.googleblog.com/2018/03/stable-channel-update-for-desktop.html

We no longer pass `use_gconf=false` to GN because the option was removed
upstream (along with all the GConf code).

We've had to backport several commits; GCC support has been deteriorating
upstream now that it is no longer officially tested. Huge thanks to @jdapena
for his work there.

Added license files:
* ios/third_party/motion_animator_objc/LICENSE: Apache 2.0. Unused in the
  Linux build.
* ios/third_party/motion_transitioning_objc/LICENSE: Apache 2.0. Unused in
  the Linux build.
* third_party/boringssl/src/third_party/fiat/LICENSE: MIT.
* third_party/libovr/LICENSE: Oculus SDK license. Unused in the Linux build.
  License: <https://developer.oculus.com/licenses/sdk-3.4.1>
* third_party/perfetto/MODULE_LICENSE_APACHE2: Apache 2.0.

Changed license files:
* third_party/content_shell_fonts/LICENSE: SIL OPEN FONT LICENSE, GPL v2,
  Bitstream Vera Fonts Copyright. Only used for testing. Changes:
  <https://chromium-review.googlesource.com/836549>
* third_party/ffmpeg/CREDITS.chromium: Unchanged. Changes:
  <https://chromium-review.googlesource.com/c/chromium/third_party/ffmpeg/+/862190>
* third_party/freetype/src/docs/FTL.TXT: Unchanged. Changes:
  <http://git.savannah.gnu.org/cgit/freetype/freetype2.git/diff/docs/FTL.TXT?id=87ddad20076d330c4139757054d783be9b3c56ab>

Removed license files:
* base/third_party/icu/LICENSE: no longer shipped to avoid duplication.
  The license is already present in third_party/icu/LICENSE.
  <https://chromium-review.googlesource.com/854838>

Test-built with:
* Yocto pyro, MACHINE="raspberrypi3" and GCC 5 and 6.
* Yocto rocko, MACHINE="intel-corei7-64", MACHINE="qemuarm64" with GCC 7.

chromium-ozone-wayland still configures fine with the new gn-native.

PR #156, opened 2018-08-09
chromium-ozone-wayland: Update to 70.0.3509.0

msisov:
Hello,

This is a chromium-ozone-wayland m70 release. It's based on a design without mus aka ui service (which we dropped in April). Many features had to be supported again and we had to adapt patches + write new ones. 

When using this recipe, users may experience some unexpected failures and crashes. It can happen due to a reason of using a dev chromium channel or wayland support. If users experience any issues, they should submit them to github.com/igalia/chromium/issues

This build was tested with yocto pyro, GCC 6 and MACHINE="raspberrypi3".

There is also a branch for Renesas R-Car M3 board, which is our target platform, available from
github.com/igalia/meta-browser.


rakuco:
Hmm, this is a big change and I've got some comments (I still need to finish reviewing this though). Please wait before merging this one.

otavio:
The `nodejs` recipe does not belong here. It should go to `meta-oe` as we already depend on it.

twoerner:
meta-oe already has a nodejs recipe:
https://layers.openembedded.org/layerindex/recipe/4427/

recipes-browser/chromium/chromium-gn.inc
```diff
@@ -9,7 +9,6 @@ OUTPUT_DIR = "out/Release"
 B = "${S}/${OUTPUT_DIR}"
 
 SRC_URI += " \
-        file://v8-qemu-wrapper.patch \
```
rakuco: This change looks unnecessary. The existing patch in `files/` also works with M70 AFAICS.

recipes-browser/chromium/chromium-gn.inc
```diff
@@ -27,6 +26,7 @@ DEPENDS += " \
     glib-2.0 \
     gn-native \
     gperf-native \
+    gtk+3 \
```
rakuco: This dependency is also in the chromium-x11 recipe. If you're putting it here you should also remove it from there.

recipes-browser/chromium/chromium-ozone-wayland-tarball.inc
```diff
@@ -20,7 +20,6 @@ LIC_FILES_CHKSUM = "\
     file://${S}/buildtools/third_party/libc++abi/trunk/LICENSE.TXT;md5=8ae94dd6195890583eee15a988b6ea79 \
     file://${S}/chrome/installer/mac/third_party/bsdiff/LICENSE;md5=0dbe7a50f028269750631fcbded3846a \
     file://${S}/chrome/installer/mac/third_party/xz/LICENSE;md5=8bbde65ae0523dc878ed1ac1252d2d14 \
-    file://${S}/chrome/third_party/chromevox/LICENSE;md5=45739090b454125d851ac7bb1bbd23a2 \
```
rakuco: @otavio has asked for license changes to be explained (see #103)

recipes-browser/chromium/files/0001-Add-memory-header-when-unique_ptr-is-used.patch
```diff
@@ -0,0 +1,42 @@
+From 87054a675e7168652f65471b4f91ab7bc8da7d32 Mon Sep 17 00:00:00 2001
```
rakuco: There's no indication of the upstream status of this patch. It must be at least sent upstream if it hasn't already.

recipes-browser/chromium/files/0001-Fix-internal-compiler-error-in-convert_move.patch
```diff
@@ -0,0 +1,32 @@
+From 96bd6714a410fed0fe4a0b563dab8dd1e7cf238b Mon Sep 17 00:00:00 2001
```
rakuco: See #135 (again I wonder why GCC 6 was targeted in your tests). This change works, but GCC emits a bunch of instructions instead a single `REV16` opcode.

recipes-browser/chromium/files/0001-IME-for-Platform-integration-remove-chromeos-deps.patch
```diff
@@ -0,0 +1,291 @@
+From a92527ed86280c666cb74b5cec1823d50fd807dc Mon Sep 17 00:00:00 2001
```
rakuco: This is quite big, and there's also no indication of the upstream status.

recipes-browser/chromium/files/0001-Use-explicitly-defined-move-constructor-with-noexcep.patch
```diff
@@ -0,0 +1,56 @@
+From 8b10822d71f0e53483bf06b8232b804a3323a170 Mon Sep 17 00:00:00 2001
```
rakuco: No upstream status or error message. Looks very similar to e.g. https://chromium-review.googlesource.com/1107918 (i.e. this fix is working around the underlying issue instead of fixing it).

recipes-browser/chromium/files/0002-IME-for-Wayland-Ozone-Wayland-IME-integration.patch
```diff
@@ -0,0 +1,1468 @@
+From 5a63b71d5ae9b73d59d6ba944e78d4aaef070b7f Mon Sep 17 00:00:00 2001
```
rakuco: No upstream information for such a big patch.

msisov:
The recipe is based on the work-in-progress branch at igalia's chromium fork. We are constantly upstreaming the patches. That's first. Secondly, the branch is rebased every week against ToT and it's not based on any tags.

recipes-browser/chromium/files/0001-Fix-internal-compiler-error-in-convert_move.patch
msisov: As I said we have a completely different target platform. That's pyro with gcc 6 was used.

recipes-browser/chromium/files/0001-IME-for-Platform-integration-remove-chromeos-deps.patch
msisov: it's downstream at the moment. And I prepared this after a tarball was created. TBH, the tarball contains around ~20 patches, which are not upstreamed. Do you want me to describe all of them? I don't think it's a good idea.

recipes-browser/chromium/files/0001-Use-explicitly-defined-move-constructor-with-noexcep.patch
msisov: we are not able to fix our target platform. Thus, fixing it here. If you think the meta-brawser at OSSystems doesn't require this, I can remove it.

recipes-browser/chromium/files/0002-IME-for-Wayland-Ozone-Wayland-IME-integration.patch
msisov: check comment above.

recipes-browser/chromium/files/0001-Add-memory-header-when-unique_ptr-is-used.patch
msisov: of course, sir. It will be sent.

msisov:
Let me close this then and prepare a new pull request based on the targets you use for chromium-x11

recipes-browser/chromium/files/0001-Add-memory-header-when-unique_ptr-is-used.patch
rakuco: FTR, I ended up fixing those two upstream yesterday

recipes-browser/chromium/files/0001-Fix-internal-compiler-error-in-convert_move.patch
rakuco: I understand it, my point was that you're better off fixing GCC 6 in pyro with the patch I sent to oe-core, or applying this one here only on older GCC versions to avoid pessimizing the code for all compilers.

recipes-browser/chromium/files/0001-IME-for-Platform-integration-remove-chromeos-deps.patch
rakuco: You don't really need to write a huge description of what's going on, I'm more interested in whether the patch is a backport/has been sent upstream/was rejected/etc.

recipes-browser/chromium/files/0001-Use-explicitly-defined-move-constructor-with-noexcep.patch
rakuco: The underlying issue is also in Chromium, so there should be no need to change the platform -- it's just that you're treating the symptoms rather than fixing the root cause (which is adding `noexcept` to the types used inside those 2 types).

rakuco:
Thanks for the work, and I hope you did not get discouraged. Was there a reason to add the recipe here if the Igalia weekly snapshots are also being published? From what I understood, it seems to make sense to update the recipe here with a more stable release once there is one.

PR #158, opened 2018-08-16
Firefox 52.9esr

ashie:
Upgrade firefox & firefox-l10n to 52.9.0.

I've confirmed that firefox 52.9.0esr can work normally on poky master@028a292001f64ad86c6b960a05ba1f6fd72199de


kraj:
have you validated it on master ?

ashie:
> have you validated it on master ?

Yes, but not the latest at this moment.
1 week behind:

> I've confirmed that firefox 52.9.0esr can work normally on poky master@028a292001f64ad86c6b960a05ba1f6fd72199de

I'll reconfirm it against latest poky & meta-openembedded.
I've noticed that the following issue should be fixed at least:

```
WARNING: firefox-l10n-ja-52.9.0esr-r3 do_populate_lic: ${COREBASE}/LICENSE is not a valid license file, please use '${COMMON_LICENSE_DIR}/MIT' for a MIT License file in LIC_FILES_CHKSUM. This will become an error in the future
```

PR #165, opened 2018-09-02
chromium: Refactor clang patch to fix chromium-ozone-wayland

kraj:
zlib fix is specific to chromium-68, so when ozone port is brought over
to 68+ then we will need to apply that part as well for now its only x11
port which needs it


kraj:
can someone comment on this if all is ok please merge it.

PR #179, opened 2018-10-03
chromium: Check for toolchain being clang as well for deciding on is_…

kraj:
…clang

There are scenarios where someone may setup TOOLCHAIN = "clang" in
local.conf but then not have meta-clang in bblayers.conf which would
actually be ignored if meta-clang is not there, so we also need to make
sure that we have 'toolchain-clang' in OVERRIDES which is only added by
meta-clang and when we want to use clang as default compiler


kraj:
@rakuco that will work too and will be more accurate match

PR #198, opened 2018-11-23
firefox: Improve building stylo

ashie:
Ensure to set better BINDGEN_CFLAGS:

* Set proper target name
* Pass "--sysroot" option instead of adding "-I" flags
* Guess clang's include path more properly
* Add "-mfloat-abi=hard" option if it's needed
  The target triplet may not have "hf" suffix even if it supports
  hardware float.

It fixes following issues:

* configure fails when clang isn't installed
* Can't buld against aarch64 when `PACKAGECONFIG_append_pn-firefox = " stylo "` is specified.


recipes-browser/firefox/firefox/fixes/0001-Add-clang-s-include-path-on-cross-compiling.patch
```diff
@@ -0,0 +1,186 @@
+From c47a19b69e6dbe82a62dd048d5c5fcf5ba603e9d Mon Sep 17 00:00:00 2001
+From: Takuro Ashie <ashie@clear-code.com>
+Date: Tue, 20 Nov 2018 09:35:37 +0900
+Subject: [PATCH] Add clang's include path on cross compiling
+
+It seems that clang doesn't include the path of its headers on
+cross compiling at least on Ubuntu-16.04. This patch ensure to
+add it.
+
```
otavio: Upstream-Status: <...> please.

ashie: Added.

otavio:
Thank you for fixing it; really appreciate :-)

PR #200, opened 2018-12-03
firefox-l10n: Fix extension filename extraction

kraj:
Current logic falls short for langpacks containing 'id' in its name the
reason is that it colludes with the sed expression used to extract the
xpi extension name from manifest. This results in errors like

Packing langpack-id@firefox.mozilla.org
Packed XPI file. It is located in
TOPDIR/build/tmp/work/all-yoe-linux/firefox-l10n-id/60.1.0esr-r3
install: target '.xpi' is not a directory

Convert this to a python function and use the json python module to
parse the manifest.json and extract the right key-value pair


kraj:
fixes error like these http://errors.yoctoproject.org/Errors/Details/202960/

PR #202, opened 2018-12-05
firefox: Fix build break after rust dependency was added

kraj:
When we added dependency on rust from meta-rust, firefox failed to build
for arm platforms since we needed to make firefox configury to
understand how to use the rust compiler and rust build environment that
is provided.

Firefox 60 builds fine with thumb2 ISA, therefore use arm ISA only for
arm <= v5 arches which uses thumb1

Use unwind model for rust panic mode, default is to use abort mode

### DIFF
--- a/classes/mozilla.bbclass
+++ b/classes/mozilla.bbclass
@@ -27,7 +27,9 @@ export HOST_AR = "${BUILD_AR}"
 
 mozilla_run_mach() {
 	export SHELL="/bin/sh"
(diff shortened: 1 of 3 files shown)


kraj:
Fixes build breaks like

http://errors.yoctoproject.org/Errors/Details/203403/

ashie:
I've confirmed that it's also fixes the build issue with meta-rust:master on my arm (Renesas RZ/G1) & arm64 (Renesas R-Car Gen3) build environment. Thanks!

PR #209, opened 2019-01-09
chromium-ozone-wayland: Update to stable 71.0.3578.98

msisov:
This update is based on the stable M71 version with
backported Ozone/Wayland patches from the ToT.

The Chromium with Ozone/Wayland recipe shares the
chromium-upstream-tarball now, and will be updated
as soon as new version is shipped.

The Chromium with Ozone/Wayland has the following
features possible to be used now -

1) Ime, when --enable-ime-service and --enable-wayland-ime
are passed as command line arguments.

2) GPU native memory buffers + zero-copy - the browser
must be run in a multi process mode (without --in-process-gpu
flag, for example), and use available drm render nodes to
create gbm buffers to draw to. Flags needed -
--enable-native-gpu-memory-buffers.

3) Partial swap - check the about://gpu page, and
if disable_post_sub_buffers_for_onscreen_surfaces
applied, pass the --disable_post_sub_buffers_for_onscreen_surfaces=0
as command line argument to enable partial swap.

4) Touch screen - for a better support pass
--touch-events=enabled

5) The browser supports all (or almost all, testings
are needed) user driven interactions including resize/drag
and etc.

6) Clipboard support is still under development and going to
be shipped soon.

7) Drag and drap has partial support only dragging is supported.

8) Tab drag is under development and will be shipped soon.

9) Proper multi-screen support is under development as well.

10) HW accelerated video decoding - we have included
v4l2 patches, which enable v4l2 device to be used for hw
accelerated video decoding on Linux machine. This code path
can be enabled with the following setting -
PACKAGECONFIG_pn-chromium-ozone-wayland = " use-linux-v4l2"

Stay tuned for other changes, and wait for new updates, which
will be more frequent from now on.

Test-built with:
 * Yocto sumo, GCC 7 and MACHINE="raspberrypi3"


rakuco:
Lots of great news there :) I'll try to test-build this patch with more configurations and review the PR until the end of the week.

kraj:
have you tested it on aarch64 ?

recipes-browser/chromium/chromium-ozone-wayland_71.0.3578.98.bb
```diff
@@ -0,0 +1,53 @@
...
+
+DEPENDS += "\
+        libxkbcommon \
+        virtual/egl \
+        wayland \
+"
+
+# Chromium can use v4l2 device for hardware accelerated video decoding. Make sure that
+# /dev/video-dec exists.
+PACKAGECONFIG[use-linux-v4l2] = "use_v4l2_codec=true use_v4lplugin=true use_linux_v4l2_only=true"
```
rakuco: Can you document this knob in `README`?

recipes-browser/chromium/chromium-gn.inc
```diff
@@ -9,6 +9,10 @@ OUTPUT_DIR = "out/Release"
 B = "${S}/${OUTPUT_DIR}"
 
 SRC_URI += " \
+        file://0001-Move-the-atspi2-target-to-a-separate-BUILD.gn-file.patch \
```
rakuco: Can you keep this list sorted?

recipes-browser/chromium/chromium-ozone-wayland_71.0.3578.98.bb
```diff
@@ -0,0 +1,53 @@
+require chromium-upstream-tarball.inc
+require chromium-gn.inc
+
+SRC_URI += " \
+        file://0001-Reland-minigbm-define-GBM_BO_IMPORT_FD_MODIFIER.patch;patchdir=third_party/minigbm/src \
```
rakuco: Any particular reason not to just adjust the patch itself rather than specifying `patchdir` here?

recipes-browser/chromium/chromium-ozone-wayland/0016-Add-support-for-V4L2VDA-on-Linux.patch
```diff
@@ -0,0 +1,298 @@
+Upstream-Status: Inappropriate
```
rakuco: Can you explain why this is inappropriate? It's also unclear how much of an effort it's going to be to continue rebasing a patch like this on every release.

recipes-browser/chromium/chromium-ozone-wayland/0017-Add-mmap-via-libv4l-to-generic_v4l2.patch
```diff
@@ -0,0 +1,68 @@
+Upstream-Status: Inappropriate
```
rakuco: Same here about why this is inappropriate upstream.

recipes-browser/chromium/chromium-ozone-wayland/0017-Add-mmap-via-libv4l-to-generic_v4l2.patch
```diff
@@ -0,0 +1,68 @@
...
+
+Signed-off-by: Maksim Sisov <msisov@igalia.com>
+---
+From 99195af8c41a7efac1527a267f32b23d9f274725 Mon Sep 17 00:00:00 2001
+From: Damian Hobson-Garcia <dhobsong@igel.co.jp>
+Date: Wed, 21 Mar 2018 13:18:17 +0200
+Subject: [PATCH 17/17] Add mmap via libv4l to generic_v4l2_device
+
+Issue #437
```
rakuco: This and the previous patch mention this issue, but it's not clear what bug tracker this is referring to (or where these patches are coming from).

msisov:
> Are all the patches there necessary? I see some patches just renaming files or flags rather than implementing new functionality or fixing bugs.

I would like to keep all of them for consistency.

> The build on qemuarm64 + thud fails to link:

Looks like we need to switch to system libdrm as well. I'll do so.

recipes-browser/chromium/chromium-ozone-wayland/0016-Add-support-for-V4L2VDA-on-Linux.patch
rakuco: Please add an explanation to the patch then, preferably with upstream discussions? And have you guys kept this patch in the tree for long? My main concern from a maintainer perspective is that M72 will come out and suddenly there's a patch that no longer applies and that I don't know how to update blocking the entire update.

recipes-browser/chromium/chromium-ozone-wayland/0017-Add-mmap-via-libv4l-to-generic_v4l2.patch
rakuco: Can you add a pointer to https://github.com/Igalia/chromium/issues/437 and hopefully a bigger explanation (as I alluded to in the other comment)? I took a look at that GitHub issue and there's just a list of things that needed to be done, so anyone not tracking that repository closely will get confused when looking at these patches.

recipes-browser/chromium/chromium-ozone-wayland/0016-Add-support-for-V4L2VDA-on-Linux.patch
msisov: We have been maintaining it since m65, if I remember right

otavio:
Thank you @rakuco! My pleasure! :+1: 

@msisov thanks for all your commitment on keeping this up to date, it is really appreciated.

PR #216, opened 2019-02-06
chromium: Update to 72.0.3626.81

msisov:
This updates chromium-x11 and chromium-ozone-wayland recipes to 72.0.3626.81

Release announcement:
https://chromereleases.googleblog.com/2019/01/stable-channel-update-for-desktop.html

chromium-ozone-wayland test-built with:
 * Yocto sumo, GCC 7 and MACHINE="raspberrypi3"
 * Yocto sumo, GCC 7 and MACHINE="intel-corei7-64"
 * Yocto Rocko, GCC 7 and MACHINE="aarch64"

chromium-x11 test-built with:
* rocko, GCC 7 and MACHINE="qemuarm64"
* sumo, GCC 7, MACHINE="raspberrypi3" and TOOLCHAIN_pn-chromium-x11="clang"
* thud, GCC 8, MACHINE="intel-core-i7-64" and MACHINE="intel-core2-32"

Ffter https://crrev.com/c/1278470, closure_compile=false
is used to avoid pulling java dependencies into the recipe.

chromium-ozone-wayland has the following patches backported:
 * Enable Clipboard with ozone,
 * Implement WaylandScreen,
 * Fix event routing when two or more windows created,
 * Fix window bounds on initial start,
 * And etc.

Signed-off-by: Maksim Sisov <msisov@igalia.com> and 
Raphael Kubo da Costa <raphael.kubo.da.costa@intel.com>

License changes:
Changed:
- buildtools/third_party/libc++abi/trunk/LICENSE.TXT - Update copyright year to 2018
- buildtools/third_party/libunwind/trunk/LICENSE.TXT - Update copyright year to 2018

Removed:
- third_party/WebKit/LICENSE_FOR_ABOUT_CREDITS - Removed with the entire directory
- tools/fuchsia/fidlgen_js/third_party/enum34/LICENSE - Removed with the entire directory.

PR #218, opened 2019-02-11
chromium: Remove unused patches

rakuco:
Drop a few patches that were used by the chromium-ozone-wayland before its
update to M71.


rakuco:
@otavio @msisov 
This is some clean-up work while I was finishing the chromium-x11 work I mentioned in #216 (it can also be merged independently of that PR)

PR #219, opened 2019-02-12
chromium: fix build with master on arm

kraj:


rakuco:
What ARM versions were broken with Chromium before? AFAIR, nothing prior to ARMv7 is supported upstream.

kraj:
> What ARM versions were broken with Chromium before? AFAIR, nothing prior to ARMv7 is supported upstream.

i added v5 for sake of completion, we mark recipe as incompatible for arm < v6 already.

rakuco:
Hmm, then I'm confused. I always test ARM and ARM64, so what `MACHINE` value is currently broken?

kraj:
> Hmm, then I'm confused. I always test ARM and ARM64, so what `MACHINE` value is currently broken?

all arm machines which inherit cortex tunes after this commit in oe-core
https://git.openembedded.org/openembedded-core/commit/?id=49bc3eab2897307a92b538a045e3d4dce245b0e0

We can no longer operate on `-march=` since this is no longer available on compiler cmdline.

rakuco:
I see. Sorry that I didn't have time to look at this today, I'll try to do that tomorrow.

rakuco:
I tried applying the same OE change to my sumo branch and `MACHINE="raspberrypi3"`. That `MACHINE` configuration has `DEFAULTTUNE ?= "cortexa7thf-neon-vfpv4"`, so I'd expect it to be affected. The final `ARM_ARCH` value remained the same though: `armv7ve`. Does your change really affect ARMv7 and up? What `MACHINE` can I use to test this?

Also, not passing `arm_arch` to GN is wrong: if you just pass `arm_version`, it will [infer an `arm_arch` value](https://chromium.googlesource.com/chromium/src/+/72.0.3626.109/build/config/arm.gni#74) that is often not the value we really want to pass to the compiler.

kraj:
This change is only needed when OE-Core master is used and is not valid for older branches, I hope you chose oe-core/poky master branch while testing. does core/master work for your to build chromium without this change?

Secondly, passing specific arm_arch is covered with `-mcpu` option where compiler will compute
a suitable value for `-march` from `-mcpu` automatically so its not required to be set.

rakuco:
Right, I've finally managed to reproduce the problem with master. You should have pointed to https://git.openembedded.org/openembedded-core/commit/?id=ac83d22eb5031f7fdd09d34a1a46d92fd3e39a3c instead, this is the commit that drops `-march` and explains what's going on.

> Secondly, passing specific arm_arch is covered with `-mcpu` option where compiler will compute
> a suitable value for `-march` from `-mcpu` automatically so its not required to be set.

I get that, what I'm saying is that Chromium's build system will pass `-march` to the compiler by default even if OE passes `-mcpu`. It [guesses a value](https://chromium.googlesource.com/chromium/src/+/72.0.3626.109/build/config/arm.gni#74) and [uses it here](https://chromium.googlesource.com/chromium/src/+/72.0.3626.109/build/config/compiler/BUILD.gn#803), so we may end up with an invalid or suboptimal combination of `-mcpu` and `-march` in the end.

Also, is the proposed change here backwards-compatible? It's not clear to me if older branches set `-mcpu` properly without `-march` too.

> i added v5 for sake of completion, we mark recipe as incompatible for arm < v6 already.

While here, can you drop the v5 and v6 lines in the final patch?

kraj:
> Right, I've finally managed to reproduce the problem with master. You should have pointed to https://git.openembedded.org/openembedded-core/commit/?id=ac83d22eb5031f7fdd09d34a1a46d92fd3e39a3c instead, this is the commit that drops `-march` and explains what's going on.
> 
> > Secondly, passing specific arm_arch is covered with `-mcpu` option where compiler will compute
> > a suitable value for `-march` from `-mcpu` automatically so its not required to be set.
> 
> I get that, what I'm saying is that Chromium's build system will pass `-march` to the compiler by default even if OE passes `-mcpu`. It [guesses a value](https://chromium.googlesource.com/chromium/src/+/72.0.3626.109/build/config/arm.gni#74) and [uses it here](https://chromium.googlesource.com/chromium/src/+/72.0.3626.109/build/config/compiler/BUILD.gn#803), so we may end up with an invalid or suboptimal combination of `-mcpu` and `-march` in the end.

thats possible, I think we have to stop chromium from guessing wrong -march and maybe remove the
code which adds -march especially for arm since we do not need it for OE. I guess we can set `arm_arch` explicitly in recipe

> 
> Also, is the proposed change here backwards-compatible? It's not clear to me if older branches set `-mcpu` properly without `-march` too.

they do. older branches set both -march/-mcpu which in itself is a problem thats what I have tried to fix with the core changes.

> 
> > i added v5 for sake of completion, we mark recipe as incompatible for arm < v6 already.
> 
> While here, can you drop the v5 and v6 lines in the final patch?

sure

kraj:
ping ?

rakuco:
Er, I thought you were going to work on those things I mentioned above and send a new version.

kraj:
Yes that’s already considered in the patch merged did you look into it ?

rakuco:
> Yes that’s already considered in the patch merged did you look into it ?

Nope, there were no notifications or indications that you were working on the PR, so when you sent the ping I thought you were just asking for an update from our/my side rather than for a new review.

PR #223, opened 2019-02-22
chromium: Fix chromium build with musl again

kraj:
Enable world build for chromium/musl again

### DIFF
--- a/recipes-browser/chromium/chromium-gn.inc
+++ b/recipes-browser/chromium/chromium-gn.inc
@@ -19,6 +19,23 @@ SRC_URI += " \
         file://do-not-specify-march-on-arm.patch \
 "
 
+SRC_URI_append_libc-musl = "\
+        file://musl/musl-include-fcntl.h-for-loff_t.patch \
+        file://musl/musl_no_mallinfo.patch \
+        file://musl/musl_no_execinfo.patch \
+        file://musl/musl_TEMP_FAILURE_RETRY.patch \
+        file://musl/musl_fix-stack.patch \
+        file://musl/musl_breakpad.patch \
+        file://musl/musl_fontconfig.patch \
+        file://musl/musl_secure_getenv.patch \
+        file://musl/musl_libc_malloc.patch \
+        file://musl/musl_gnu_libc-version.patch \
+        file://musl/musl_resolver.patch \
+        file://musl/musl_off64_t.patch \
+        file://musl/musl_lss-match_syscalls.patch \
+        file://musl/musl_crashpad.patch \
+        file://musl/musl_replace_libc_fpstate.patch \
+"
 # This patch could be enabled for clang builds, one just needs to make it work.
 # On sumo and MACHINE="raspberrypi3", the build fails with errors such as
 #   /tmp/crc32_simd-01be44.s:86: Error: selected processor does not support `crc32b r0,r0,r5' in Thumb mode
@@ -257,6 +274,11 @@ ARM_VERSION_armv7a = "7"
 ARM_VERSION_armv7ve = "7"
 ARM_VERSION_armv6 = "6"
 
+# GN computes and defaults to it automatically where needed
+# forcing it from cmdline breaks build on places where it ends up
+# overriding what GN wants
+TUNE_CCARGS_remove = "-mthumb"
+
 GN_ARGS_append_arm = ' \
         arm_float_abi="${ARM_FLOAT_ABI}" \
         arm_fpu="${ARM_FPU}" \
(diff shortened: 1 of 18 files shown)


recipes-browser/chromium/chromium-gn.inc
```diff
@@ -19,6 +19,22 @@ SRC_URI += " \
...
+        file://musl/musl_fix-stack.patch \
+        file://musl/musl_breakpad.patch \
+        file://musl/musl_fontconfig.patch \
+        file://musl/musl_secure_getenv.patch \
+        file://musl/musl_libc_malloc.patch \
+        file://musl/musl_gnu_libc-version.patch \
+        file://musl/musl_resolver.patch \
+        file://musl/musl_off64_t.patch \
+        file://musl/musl_lss-match_syscalls.patch \
+	file://musl/musl_crashpad.patch \
```
otavio: Bad indent, please fix.

recipes-browser/chromium/chromium-gn.inc
```diff
@@ -257,6 +273,8 @@ ARM_VERSION_armv7a = "7"
 ARM_VERSION_armv7ve = "7"
 ARM_VERSION_armv6 = "6"
 
+TUNE_CCARGS_remove = "-mthumb"
```
rakuco: Why?

recipes-browser/chromium/chromium-gn.inc
```diff
@@ -270,6 +288,9 @@ GN_ARGS_append_armv6 += 'use_allocator="none"'
 # https://bugs.chromium.org/p/webrtc/issues/detail?id=6574
 GN_ARGS_append_armv6 += 'arm_use_neon=false'
 
+# Disable glibc shims on musl
```
rakuco: The comment removed in #106 explained this in more detail.

otavio:
As @kraj is actively maintaining those patches, I am not against to keep them. He may test musl when new releases are sent and update those patches. 

@kraj what you think?

kraj:
Yes I plan to maintain them so no issues there and moreover they are only applied conditionally so no one else has to worry about it if they don't want to. Regarding upstreaming yes I agree, a lot of support is already accepted in chromium,  This will provide additional vetting and upstreaming from here. However it can take a while since the patches are not considering non-linux platforms in many cases.

recipes-browser/chromium/chromium-gn.inc
kraj: -mthumb is computed by chromium build system and is used by default, so one its redundant anf second its problematic because, there are certain pieces in chromium which are not buildable in thumb mode and chromium build system switched to use arm mode for those pieces only, forcing ISA globally like this breaks builds on those particular subsystems
especially with clang, who is more picky on assembler errors ( e.g. use of LDREX in thumb mode )

recipes-browser/chromium/chromium-gn.inc
kraj: I think comment is valuable, I will  add it along.

recipes-browser/chromium/chromium-gn.inc
kraj: sure

kraj:
All comments are addressed in the new version I just pushed

kraj:
> Even if @kraj keeps maintaining those patches here for the foreseeable future, this means we will have:
> 
> * chromium-x11 + glibc (currently tested before any update to the recipe)
> * chromium-ozone-wayland + glibc (currently tested before any update to the recipe)
> * chromium-x11 + musl (untested, can break with any update)
> * chromium-ozone-wayland + musl (untested, can break with any update)
> 

thats ok

> and I don't see this situation changing: none of the patches being added have been upstreamed, and they were not upstreamed in the past either. In fact, if I `git log --grep=musl` in Chromium's main repository the latest commit was actually made by me in late 2017 when I tried to get some of the patches we had lying around upstream.
> 

you assume meta-browser is just for whats maintained upstream, but actually its a layer for browsers
for openembedded and we can carry patches.

> I still believe there isn't enough reason to support musl in this repository, especially with no actual plan of working with the Chromium developers on keeping things working upstream as much as possible. At best, we've got patches that neither @msisov nor myself maintain for what (IMO, no actual data to back this up) looks like a niche group, at worst we're touching Chromium's internals without any review by Chromium's developers and potentially breaking things in subtle ways.

it has to start somewhere, 
musl is a standard alternative in OE-core and isnt that enough for a extra layer to support it ?

recipes-browser/chromium/chromium-gn.inc
kraj: its not musl specific but armv7 specific, it was just that it was found during musl build

otavio:
As @kraj will be doing the tests and rework, I am fine in merging it. @rakuco and @msisov, you guys don't need to test musl builds as @kraj will be doing so.

@kraj mind to rebase?

kraj:
Done

PR #233, opened 2019-03-15
Only enable lld when set in DISTRO_FEATURES

MastaG:
When using meta-clang, one can use the lld compiler by setting ld-is-lld in DISTRO_FEATURES.
Verified on arm 32bit using master branches of meta-clang, openembedded-core and meta-openembedded.


otavio:
I am fine with the change however please fix the commit log. It must have recipe name as prefix.

MastaG:
like this?

kraj:
yes however, this is a include file which is used in both x11 and wayland versions. So you might want to say

```
chromium-ozone-wayland,chromium-x11: enable lld when set in distro features
```

rakuco:
When I'm doing something that touches both recipes, I just prefix it with "chromium" instead.

MastaG:
Alright , I'll change it to chromium once I get home :)

MastaG:
I now only prefixed it with: chromium:

PR #236, opened 2019-03-18
chromium: Fix do_configure() after 7c6d7bb

rakuco:
Commit 7c6d7bb ("chromium: enable lld when set in distro features") broke
do_configure() as instead of passing a key=value pair to GN it was just
passing the value half.

Spotted by @diego-santacruz.


rakuco:
Small hotfix, merging.

PR #240, opened 2019-03-20
chromium: minigbm doesn't seem to link statically

MastaG:
Even when building non-component builds, libminigbm.so isn't linked in statically.
So let's check if the library exists and ship it anyways.

### DIFF
--- a/recipes-browser/chromium/chromium-gn.inc
+++ b/recipes-browser/chromium/chromium-gn.inc
@@ -430,6 +430,13 @@ do_install() {
 		install -m 0755 *.so ${D}${libdir}/chromium/
 	fi
 
+	# When building chromium with use_system_minigbm=false,
+	# libminigbm.so does not seem to get linked in statically.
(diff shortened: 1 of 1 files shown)


msisov:
Please add to the comment that it only happens when building with use_system_minigbm=false

MastaG:
like this?

recipes-browser/chromium/chromium-gn.inc
```diff
@@ -430,6 +430,13 @@ do_install() {
 		install -m 0755 *.so ${D}${libdir}/chromium/
 	fi
 
+	# When building chromium with use_system_minigbm=false,
+        # libminigbm.so does not seem to get linked in statically.
```
msisov: nit: remove space to make the comment vertically aligned :)

msisov:
Awesome!

otavio:
Please next time, just make sure to have the `Signed-off-by.

PR #244, opened 2019-03-28
firefox: Remove "installdir"

ashie:
Because it doesn't take effect.
"--libdir" is already added to EXTRA_OECONF instead.


otavio:
Ok, merged. But did you fix the building on multilib environment?

ashie:
Yes, I uses same environment with @cosmo0920.

PR #260, opened 2019-05-14
chromium-ozone-wayland: Fix usage without zwp_linux_dmabuf and NGMB.

msisov:
This commit fixes two issues:
1) When the zwp_linux_dmabuf protocol is not available, the
WaylandConnection still has been trying to access the manager
and add a Window to it. This patch avoids this behaviour.

2) NativeGpuMemoryBuffers support was broken after a major refactoring
done. Now, it works again.


rakuco:
looks sane to me

MastaG:
I can confirm that this fixes wayland-egl support on the Odroid XU3 using the ARM mali r12p004 wayland drivers.
It now runs with and without --in-process-gpu.

Issue #237, opened 2019-03-18
chromium: Decide which toolchain(s) to support


rakuco:
So far, we've supported building Chromium (both chromium-x11 and chromium-ozone-wayland) with whatever GCC (or clang, via meta-clang) was in a given Yocto/OE branch.

This is becoming increasingly difficult:
* GCC is not officially supported upstream, which means @msisov and I need to spend a considerable amount of time submitting build fixes upstream. Some of them are actual code fixes, but most of them are just workarounds for GCC bugs, and depending on the kind of workaround they are not very welcome upstream (and sometimes are rejected altogether). This is becoming untenable: fixing Chromium is time-consuming, but reporting bugs and reaching a wall in GCC's Bugzilla is frustrating. And even when the bugs do get fixed there's the problem of getting the fixes to the users of stable GCC versions. Additionally, I know that there are some clang-specific ifdefs in the Chromium code that leverage some builtins that can offer performance improvements that are just not used with GCC.
* clang via meta-clang is a better alternative, but being stuck with an old (by Chromium's standards) version in a branch doesn't help either. For example, I'm testing an update to M73 with @msisov and right now the code just doesn't build with clang6 in the sumo branch. This was fixed in clang7, but cherry-picking the revision that fixed things isn't enough (I reach an ICE if I do that), and adding some upstream-inappropriate patch to work around the issue can be difficult and result in new bugs or performance regressions.

Officially, Chromium only really supports building a certain release with a specific LLVM/clang revision. There are prebuilt binaries that can be downloaded, but they can also be built locally. There are no Chromium-specific patches, Chromium's clang is essentially a snapshot from LLVM's svn repository.

I'd like to hear what your opinions are here. If we were to follow upstream's advice, we'd build a copy of Chromium's clang as a native recipe and it to build the code (I have no idea how this would work in terms of all the toolchain-related variables that get set in e.g. `bitbake.conf` though).

rakuco:
@otavio @kraj @msisov

otavio:
There are a few hidden things here:

1. number of branches
2. GCC
3. clang

I think we should reduce the number of branches we support instead. I see no reason for `master` to be tested against `sumo` for example. In fact, we could test it on `thud` and `master` and it would be enough, from my point of view.

GCC vs clang is more difficult from my perspective. The biggest problem in "dropping" GCC support is it will add a hard dependency on `meta-clang`.

[...]

PR #264, opened 2019-05-26
dynamic-layers: move Firefox to rust-layer

cpboyd:
Per https://github.com/OSSystems/meta-browser/issues/237#issuecomment-495435011

As the title states, this is moving the `firefox` recipe to `dynamic-layers/rust-layer/`.

I opted not to move the other Firefox-related recipes since they don't implicitly depend on meta-rust.

Granted, **firefox-l10n.inc** and **firefox-addon.inc** contain this commented line:
`# RDEPENDS_${PN} = "firefox"`

If it's uncommented, they'll need to be moved to the **rust-layer** folder


rakuco:
cc @otavio @kraj, I'm not familiar with dynamic layers and I guess you guys are the best to comment on this

kraj:
@cpboyd dont they depend upon firefox ? and whats the point of having them without firefox ?

ashie:
FYI I'm now working on Firefox 68: https://github.com/webdino/meta-browser/tree/firefox-68-wip
From this version Firefox will depend on both meta-rust & meta-clang because recent firefox can't build with GCC (and I don't want to spent my time to address them).

I'm not familiar with dynamic-layers, so I'm now wondering how to solve such case with this directory structure?

cpboyd:
@kraj As I mentioned, the BitBake recipes don't currently seem to... (the `RDEPENDS` is commented out)
Perhaps for organization purposes, the other Firefox-related recipes should be moved to the rust-layer too.

@ashie Interesting point that's be related to #237.
I suppose if that's the case, then meta-clang should be a hard requirement for meta-browser (as both Chromium and Firefox require it).

otavio:
@cpboyd I'd prefer to have them all under the same namespace. Also, it seems that the recipes need to be fixed to rdepends on firefox. 

Could you add extra commits to address it?

cpboyd:
@otavio Sure, no problem.

However, with regards to the `RDEPENDS`, this commit https://github.com/OSSystems/meta-browser/commit/176d9edd1148064e94ff3fe7f957a057d3acd614 implies that the dependency on `firefox` was dropped to keep `inherit allarch`.

With that in mind, should I still revert the changes from that commit?

otavio:
We can add the ABI whitelist for it, so it does not get rebuild when firefox architecture changes.

cpboyd:
@otavio I'm not familiar with ABI whitelisting.  Is there a field to be added to the BitBake recipes dependent upon Firefox?

otavio:
> @otavio I'm not familiar with ABI whitelisting. Is there a field to be added to the BitBake recipes dependent upon Firefox?

Like this:

https://patchwork.openembedded.org/patch/144057/

cpboyd:
I've successfully rebuilt my own project, but I'm facing this issue when I try to build Firefox: https://github.com/meta-rust/meta-rust/issues/240

conf/layer.conf
```diff
@@ -9,6 +9,102 @@ BBFILES_DYNAMIC += " \
 	rust-layer:${LAYERDIR}/dynamic-layers/rust-layer/*/*/*.bbappend \
 "
 
+SIGGEN_EXCLUDERECIPES_ABISAFE_rust-layer += " \
+   firefox-addon-webconverger \
+   firefox-l10n-ach \
```
cpboyd: A potential concern here is that if more Firefox localizations are added, they'll need to be added to this list.

otavio: Agreed but it is still doable and allow for correct dependency chain.

PR #279, opened 2019-06-27
README: update list of maintainers

OnkelUlla:
Signed-off-by: Ulrich Ölmann <u.oelmann@pengutronix.de>

### DIFF
--- a/README
+++ b/README
@@ -23,13 +23,18 @@ revision: HEAD
 Contributing
 ------------
 
-To contribute to this layer you should the patches for review to the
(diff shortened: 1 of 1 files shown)


OnkelUlla:
Is the openembedded-devel list still the first place for patches and discussions? My feeling is that it is replaced by GitHub's workflow these days. Perhaps this should be reflected in the central README as well.
Please just let me know and I will readily adjust it accordingly.

otavio:
You are right; we are using GitHub for all for some time.

PR #282, opened 2019-07-03
firefox: Add a missing dependency "dbus-glib"

ashie:
I failed to build firefox without this dependency against poky:master(bc5f6725af04417dcf5c65981d8b85abee9c61d8).


ashie:
```
|  0:07.87 DEBUG: configure: error: Library requirements (dbus-glib-1 >= 0.60) not met; consider adjusting the PKG_CONFIG_PATH environment var
iable if your libraries are in a nonstandard prefix so pkg-config can find them.                                                              
|  0:07.87 ERROR: old-configure failed                                                                                                        
```

PR #284, opened 2019-07-15
Update to firefox 68.0esr

### LINKED ISSUE #265: Road to Firefox 68
According to [Firefox release Calendar](https://wiki.mozilla.org/Release_Management/Calendar), next ESR of Firefox will be shipped on 2019-07-09. So I'm now trying to port alpha or beta versions of it to Yocto.

  * https://github.com/webdino/meta-browser/tree/firefox-68-wip

Current status:

* [An alpha version already get working on AArch64](https://github.com/webdino/gecko-embedded/issues/70#issuecomment-482408084) although some additional host tools and hacks are required.
* Beta version isn't succeeded to compile yet since Rust-1.34 is required (current meta-rust is Rust-1.33).
* Most backport patches are removed since they are already merged into upstream.

Issues:

* [x] Rust-1.34 is required so we need to update meta-rust.
* [x] Node.js 8.11 or later is required as a native toolchain.
  * It's already resolved. The thud & master branch of meta-openembedded satisfies this condition.
* [x] [cbindgen](https://github.com/eqrion/cbindgen) is required but probably no layer provides it. So we need to create a new recipe.
  * I'm now using host's one.
* [x] autoconf2.13 is required as a host tool.
  * I'll add a patch to omit it as same as current master of meta-browser.
* [x] Recent FIrefox cannot build against GCC, so meta-clang will be needed.
  * I compiled the alpha version with it.
  * I'm not sure how to switch to it properly. I'm now adding the following lines to my local.conf:
```
TOOLCHAIN_pn-firefox = "clang"
AS_pn-firefox="${TARGET_PREFIX}clang"
```

ashie:
* Update to firefox 68.0esr
* Support only Clang via meta-çlang, you can't use GCC to build it
    * The dependency isn't added to conf/layer.conf yet because:
        * Chromium still uses GCC
        * meta-clang isn't required if you don't build firefox
* Add a new recipe for a toolchain "cbindgen" which is required to build modern firefox
* Update Debian's patch set
* Remove or update obsolete patches
* firefox-l10n is also updated
* Update README to add more descriptions about the recipe for firefox

Fix #265


dynamic-layers/rust-layer/recipes-devtools/cbindgen/cbindgen_0.8.7.bb
```diff
@@ -0,0 +1,82 @@
...
+# If this is git based prefer versioned ones if they exist
+# DEFAULT_PREFERENCE = "-1"
+
+# how to get cbindgen could be as easy as but default to a git checkout:
+# SRC_URI += "crate://crates.io/cbindgen/0.8.7"
+SRC_URI += "git://github.com/eqrion/cbindgen.git;protocol=https"
+SRCREV = "7b9ea991bdabb442093014d21fde6a192f574f21"
+S = "${WORKDIR}/git"
+CARGO_SRC_DIR=""
+PV_append = ".AUTOINC+7b9ea991bd"
```
otavio: I'd use:

```
PR_append = "${SRCPV}"
```

ashie: Do you mean `PV_append`?

```diff
-PR_append
+PV_append
```

ashie: done.

kraj: this `PV_append` is not needed, please remove.

README
```diff
@@ -14,7 +14,11 @@ URI: git://git.openembedded.org/meta-openembedded
 branch: master
 revision: HEAD
 
-For Firefox, meta-rust is necessary:
+For Firefox, meta-rust and meta-clang are necessary:
+
+URI: git@github.com:kraj/meta-clang
```
kraj: this should be `git://github.com/kraj/meta-clang`

dynamic-layers/rust-layer/recipes-devtools/cbindgen/cbindgen_0.8.7.bb
ashie: done.

README
ashie: done.

rakuco:
> I think we ought to move to clang on master now and stop supporting GCC as it reduces the amount of work we have to keep the recipes updated.
> 
> @rakuco @kraj what do you think?

I'll try to find some time this week to move this forward.

kraj:
@ashie  I am getting fetch errors e.g.

https://errors.yoctoproject.org/Errors/Details/252512/

for many firefox-l10n recipes

ashie:
I'll see them.

PR #290, opened 2019-07-19
chromium: Start requiring clang (and the meta-clang layer) to build

### LINKED ISSUE #237: chromium: Decide which toolchain(s) to support

rakuco:
In other words, drop GCC support altogether. Since Firefox already requires
clang, we can readd "clang-layer" to LAYERDEPENDS.

Upstream Chromium has not tested or officially supported GCC for years, and
maintaining support in the chromium-x11 and chromium-ozone-wayland recipes
has been taking more and more effort with each new major Chromium release:
- Some patches are for actual compliance issues, which we need to submit
  upstream and backport here.
- Some patches are workarounds for GCC bugs that we have to report to the
  GCC bug tracker, where they usually sit untouched for months (or maybe
  even years). Some of those patches are accepted upstream, but many are not
  upstreamable at all.
- Some patches are workarounds for old GCC bugs that have been fixed in
  newer OE/Yocto branches, but since we go as far back as "rocko" in terms
  of support right now we still need to maintain those patches around (and,
  of course, it does not make sense to upstream them).

Additionally, since upstream does not test GCC builds they tend to be less
optimized than clang ones and, in the case of https://crbug.com/941696, they
even generate wrong code.

Drop all our GCC support patches from `files/`, and also switch gn-native to
`clang-native` rather than using the system's GCC.

One consequence of this switch is that we can no longer support the "rocko"
and "sumo" branches. Upstream Chromium uses its own clang (which is a
snapshot of LLVM trunk) to build, and we cannot use any releases that are
too old. At the time of writing, with Chromium M75 in the stable channel,
this means at least clang 7.0.0 is required, which is only present in the
"thud" branch of the meta-clang layer. I tested the "rocko" and "sumo"
branches but they did not work:
- rocko (clang 5), chromium-x11 and MACHINE="qemuarm64" failed to build
  clang-native:
  `[...]/tmp/work/x86_64-linux/clang-native/5.0.1+gitAUTOINC+1368f4044e_9fc5dfce8e-r0/git/include/llvm/ExecutionEngine/Orc/OrcRemoteTargetClient.h:722:26: error: could not convert ‘((llvm::orc::remote::OrcRemoteTargetClient<ChannelT>*)this)->callB<llvm::orc::remote::OrcRemoteTargetRPCAPI::ReadMem>(Src, Size)’ from ‘Expected<vector<unsigned char,allocator<unsigned char>>>’ to ‘Expected<vector<char,allocator<char>>>’`
- sumo (clang 6), chromium-x11 and MACHINE="raspberrypi3" originally failed
  with:
  `clang-6.0: error: unknown argument: '-fcrash-diagnostics-dir=../../tools/clang/crashreports'`
  Disabling that argument caused the compiler to run into an infinite loop
  when building `third_party/skia/src/gpu/ops/GrQuadPerEdgeAA.cpp`.
- sumo (clang 6), chromium-ozone-wayland and MACHINE="beaglebone" failed
  with:
  `../../gpu/command_buffer/client/transfer_buffer_cmd_copy_helpers.h:91:37: error: constexpr variable 'checked_elements_size' must be initialized by a constant expression`
  `../../gpu/command_buffer/client/transfer_buffer_cmd_copy_helpers.h:93:17: error: static_assert expression is not an integral constant expression`

This has been test-built on:
- thud:
  - chromium-x11: `MACHINE="intel-corei7-64"` and `MACHINE="intel-core2-32"`
  - chromium-ozone-wayland: `MACHINE="qemuarm64"`
- warrior
  - chromium-x11: `MACHINE="intel-corei7-64"`
  - chromium-x11: `MACHINE="raspberrypi3"`

I also ran core-image-sato with thud and `MACHINE="intel-corei7-64"` on an
Intel NUC with an i5-72600U processor without issues.

Fixes #237.


msisov:
Does it mean we also stop sending GCC patches to Chromium? I suppose it will be only dape then

rakuco:
If they're only for this recipe, then I think so, yes. Personally I haven't done that in a while, as I think right now supporting GCC upstream is a never-ending uphill struggle (especially when the build issues are caused by bugs in GCC itself).

rakuco:
(I'm still sending patches to keep libstdc++ support working though, as we haven't switched to libc++)

msisov:
well, rs lgtm. but lets see what @kraj and @otavio say

rakuco:
* Should there be a check that causes the build to fail if one's using either rocko or sumo? How can that be done?
* Is there a need to back up master before this change for people to continue being able to build M75 with GCC? This was discussed in #237, but right now I guess oldstable is now "thud" anyway.

kraj:
@rakuco  in `conf/layer.conf` change
```
LAYERSERIES_COMPAT_browser-layer = "rocko sumo thud warrior"
```
to 
```
LAYERSERIES_COMPAT_browser-layer = "thud warrior"
```

kraj:
> * Should there be a check that causes the build to fail if one's using either rocko or sumo? How can that be done?
> * Is there a need to back up master before this change for people to continue being able to build M75 with GCC? This was discussed in #237, but right now I guess oldstable is now "thud" anyway.

maybe its worth creating a tag something like `last_gcc_compatible`

rakuco:
> @rakuco in `conf/layer.conf` change

I meant something for the chromium recipes only, as otherwise Firefox would stop working on those branches too.

kraj:
> > @rakuco in `conf/layer.conf` change
> 
> I meant something for the chromium recipes only, as otherwise Firefox would stop working on those branches too.

compatibility is per layer, and I dont think we should go too granular here. If we do above two changes we are good.

ashie:
> LAYERSERIES_COMPAT_browser-layer = "thud warrior"

Firefox already can't build against rocko & sumo because the version of nodejs-native isn't satisfied.
(It can build if you introduce it as a host tool and set `HOSTTOOLS` though)

rakuco:
Thanks, @ashie. I've pushed another commit removing those branches from `LAYERSERIES_COMPAT`, and I've adjusted the commit message for the big commit mentioning I've also tested this with warrior and `MACHINE="raspberrypi3"`.

rakuco:
Since there have been no objections, I'm going to merge this PR. @otavio, can you take care of creating a branch or tag for the commit prior to the ones I'll merge if necessary?

PR #294, opened 2019-08-05
firefox: Avoid crash on opening about:support on certain SoC

ashie:
Update the patch to disable `eglQueryString(nullptr, EGL_EXTENSIONS)`.
It causes crash on certain environments, and it's not needed on most
environments.


otavio:
I am not against this but curious if a real fix or issue is reported upstream?

ashie:
Not yet.

Because the root cause is certain proprietary GPU drivers.
But, yes, it's also worth to track at upstream. I'll report it.

At least I confirmed the bug on Renesas RZ/G1.
I reported it to Renesas long ago but I don't know that the issue is handled or not in them.

otavio:
So please report it and add the issue tracking number on the patches.

ashie:
(Bug 1571603](https://bugzilla.mozilla.org/show_bug.cgi?id=1571603)

I'll add the bug number to the patch.

ashie:
> I'll add the bug number to the patch.

done.

PR #305, opened 2019-09-18
Fix crash on ARM

### LINKED ISSUE #299: x11 in raspberry pi 4 chromium can not display on weida hi-tech cool touch screen in Yotco project
hi I open a n issue at here :
 https://github.com/agherzan/meta-raspberrypi/issues/467
maybe  the problem is from the chromium itself.
so I open at here. thank you to keep it. if I am wrong, please remove it.
I can not run chromium in raspberry pi4: ther version is chromium-x11_76.0.3809.100

 simply run : sudo -u myuser chromium in my old source code in pi3, works well.
but in current pi4 still it's Segmentation fault. I am going to add components one by one in local.conf to try.

raspberrypi4:/home/myuser# sudo -u myuser chromium  
Segmentation fault 

thanks

### LINKED ISSUE #304: chromium-x11  Segmentation fault on IMX6Q
Hi!

I'm trying to build cchromium-x11 under yocto 2.7 for the imx6q.

The application unfortunately throws a segmentation fault.

I built with -fuse-ld=gold does not make a difference.

gdb --args chromium-bin --no-sandbox shows:
Thread 1 "chromium-bin" received signal SIGSEGV, Segmentation fault.
0x75de2fbe in unw_set_reg () from /usr/lib/libc++abi.so.1

Anyone ideas?

Thanks
Dirk

kraj:
Fixes #304.
Fixes #299.

### DIFF
--- a/recipes-browser/chromium/chromium-gn.inc
+++ b/recipes-browser/chromium/chromium-gn.inc
@@ -149,8 +149,8 @@ GN_ARGS += "is_debug=false is_official_build=true"
 GN_ARGS += "use_custom_libcxx=false"
 
 # When using meta-clang, one can switch to using the lld linker
-# by using the ld-is-lld distro feature.
-GN_ARGS += "use_lld=${@bb.utils.contains('DISTRO_FEATURES', 'ld-is-lld', 'true', 'false', d)}"
+# by using the ld-is-lld distro feature otherwise use gold linker
+GN_ARGS += "${@bb.utils.contains('DISTRO_FEATURES', 'ld-is-lld', 'use_lld=true', 'use_gold=true', d)}"
 
 # By default, passing is_official_build=true to GN causes its symbol_level
 # variable to be set to "2". This means the compiler will be passed "-g2" and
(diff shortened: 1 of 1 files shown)


kraj:
> LLD linker fails:
> 
> | ld.lld: error: found local symbol 'NSS_3.12.10' in global part of symbol table in file /home/entwicklung/yocto/yocto2.7/build/tmp/work/cortexa9t2hf-neon-mx6qdl-fslc-linux-gnueabi/chromium-x11/76.0.3809.132-r0/recipe-sysroot-native/usr/lib/libsmime3.so
> | clang-8: error: linker command failed with exit code 1 (use -v to see invocation)

thanks for testing it out. What distro is running on your build host? it seems to be failing in a link for some native binary

kraj:
in v2 I have dropped the lld switch patch, I guess we will need to use clang-9 for lld to work better. which means we need to rely on master/meta-clang. the updated patch to disable llvm unwinder should fix the crash you are seing please test it out and let us know

recipes-browser/chromium/chromium.inc
```diff
@@ -11,6 +11,7 @@ S = "${WORKDIR}/chromium-${PV}"
 # requires an ever-growing amount of backports and custom patches, without any
 # guarantee that the generated code will work as expected.
 TOOLCHAIN = "clang"
+UNWINDLIB = "--unwindlib=libgcc"
```
rakuco: Can you add an empty line between this and the previous one, and add a comment explaining why this is being done and referencing #304?

kraj:
Please note that this will fix the build with meta-clang/master, I am pushing a different fix for meta-clang/warrior to llvm itslef in meta-clang which is under test and should fix the problem.

PR #309, opened 2019-09-21
Switch to gold linker as default

kraj:
This should work on thud and warrior as well. It should speed up linking

### DIFF
--- a/recipes-browser/chromium/chromium-gn.inc
+++ b/recipes-browser/chromium/chromium-gn.inc
@@ -149,8 +149,8 @@ GN_ARGS += "is_debug=false is_official_build=true"
 GN_ARGS += "use_custom_libcxx=false"
 
 # When using meta-clang, one can switch to using the lld linker
-# by using the ld-is-lld distro feature.
-GN_ARGS += "use_lld=${@bb.utils.contains('DISTRO_FEATURES', 'ld-is-lld', 'true', 'false', d)}"
+# by using the ld-is-lld distro feature otherwise use gold linker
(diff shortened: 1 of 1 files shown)


recipes-browser/chromium/chromium-gn.inc
```diff
@@ -149,8 +149,8 @@ GN_ARGS += "is_debug=false is_official_build=true"
 GN_ARGS += "use_custom_libcxx=false"
 
 # When using meta-clang, one can switch to using the lld linker
-# by using the ld-is-lld distro feature.
-GN_ARGS += "use_lld=${@bb.utils.contains('DISTRO_FEATURES', 'ld-is-lld', 'true', 'false', d)}"
+# by using the ld-is-lld distro feature otherwise use gold linker
+GN_ARGS += "${@bb.utils.contains('DISTRO_FEATURES', 'ld-is-lld', 'use_lld=true', 'use_gold=true', d)}"
```
rakuco: The logic in GN for deciding whether to use lld or gold is more intricate. Essentially, if we're using clang [`use_lld` will be set to true by default](https://cs.chromium.org/chromium/src/build/config/compiler/compiler.gni?l=189&rcl=3ee19b68cf617bc250ea03a566ace3995373aff3), so just setting `use_gold=true` [will not have the desired effect](https://cs.chromium.org/chromium/src/build/config/compiler/BUILD.gn?l=411&rcl=3ee19b68cf617bc250ea03a566ace3995373aff3).

You need to pass both `use_lld=false` and `use_gold=true` in the else case above.

kraj:
check this now

kraj:
@otavio can you review and merge it if all is ok ?

PR #311, opened 2019-09-23
chromium: Use the right bb.utils call in do_check_fdlimit_for_linking

rakuco:
`bb.utils.contains()` requires both "ld-is-gold" and "ld-is-lld" to be
present in `DISTRO_FEATURES`, which in practice means the rlimit check will
happen even when one is not using ld.bfd.

Switch to `bb.utils.contains_any()`, which tests for the presence of any of
the arguments, as we originally intended.


rakuco:
Even though we're likely getting rid of this function in #309, I'd like the code to work as expected first in case we need to revert that PR.

PR #317, opened 2019-10-08
layer: Add zeus to compatible release

kraj:


kraj:
This needs to go in ASAP, all builds will latest master are otherwise broken

PR #321, opened 2019-10-24
firefox: Import latest OpenMAX IL 1.1.2 headers

anyakichi:
OpenMAX IL 1.1.2 header files were released twice on the same version,
and Firefox 68 includes the OLD headers.

The main change between the two versions is definition of fixed-sized
integers.  For example, the old version defines OMX_U32 as unsigned long,
but the new version defines it as uint32_t.

The definition of the old version is apparently wrong, because there is
no guarantee that unsigned long is 32-bit length.  Actually unsigned
long is 64-bit length on ARM64 and this mismatch cause a problem when
using OpenMAX on RZ/G2 series SoCs.


otavio:
Please, @anyakichi , address these minor things

1. add the Upstream-Status header to the patch file
1. add your Signed-off-by to the commit

@ashie do you mind to review this one?

anyakichi:
Thank you for reviewing.

I've done requested things.  Check them, please.

diffs: https://github.com/OSSystems/meta-browser/commit/76265e33b40e84d76bc51acdbe5ef275ed7f40e2
squashed and signed: https://github.com/OSSystems/meta-browser/commit/b08c308f0fc0af390b73f068cd54d5a3a71aa867

PR #337, opened 2019-12-05
musl forward ports,pa13, clang warnings  fixes

### LINKED ISSUE #341: chromium(-x11) fails to build with pulesaudio version >= 13
With recent oe-core I get errors as

> gen/media/audio/pulse/pulse_stubs.cc:280:6: error: conflicting types for 'pa_context_errno'
> int  pa_context_errno(pa_context *c) {
>      ^

Seems pulseaudio did change their API and added some `const` here and there.
Found [1] but this won't do for us because it would break builds for pulseaudio < V13.

Anybody?

[1] https://694896.bugs.gentoo.org/attachment.cgi?id=590500

kraj:
Fixes #341.


rakuco:
Sorry for the delay, I'll try to look into it in the next few days. Can you split the musl and non-musl parts into separate PRs? The musl bits look fine from a quick glance, but the other ones need more attention.

kraj:
> Sorry for the delay, I'll try to look into it in the next few days. Can you split the musl and non-musl parts into separate PRs? The musl bits look fine from a quick glance, but the other ones need more attention.

They are already separated out as separate patches  what will separate pulls get us ?

rakuco:
It's the same old problem of landing completely independent things in a single commit: it may work, but it makes reverting more difficult, and it's more complicated to reason about a commit when it is no longer doing only one thing. And, in this specific case, it takes longer to land everything rather than just half of it.

kraj:
But commits are still independent just pull request is same how does reverting commits get impacted ?

rakuco:
My bad, I usually just look at the "Files changed" tab and didn't noticed there were multiple commits. In this case I guess only the point of landing some things faster still applies.

recipes-browser/chromium/files/64bit_time_t.patch
```diff
@@ -0,0 +1,54 @@
+Fix build on 32bit arches with 64bit time_t
+
+time element is deprecated on new input_event structure in kernel's
+input.h [1]
+
+[1] https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit?id=152194fe9c3f
+
+Upstream-Status: Pending
```
rakuco: Has the discussion with upstream started yet? I tried looking at the bug tracker and Gerrit but couldn't find anything. I think the upstream solution might end up looking a bit different, so it'd be good if you could at least file a bug with instructions for how to reproduce this issue before we land this specific patch.

kraj: No, `pending` means upstreaming activity has not started yet see [1]

[1]. https://www.openembedded.org/wiki/Commit_Patch_Message_Guidelines#Patch_Header_Recommendations:_Upstream-Status

recipes-browser/chromium/chromium-gn.inc
```diff
@@ -16,26 +16,31 @@ SRC_URI += " \
         file://wrapper-extra-flags.patch \
         file://do-not-specify-march-on-arm.patch \
         file://add_internal_define_armv7ve.patch \
+        file://Fix_building_with_pulseaudio_13.patch \
```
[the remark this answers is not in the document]
kraj: sure

recipes-browser/chromium/chromium-gn.inc
```diff
@@ -16,26 +16,31 @@ SRC_URI += " \
...
         file://musl/musl_no_execinfo.patch \
         file://musl/musl_TEMP_FAILURE_RETRY.patch \
         file://musl/musl_fix-stack.patch \
         file://musl/musl_breakpad.patch \
         file://musl/musl_fontconfig.patch \
-        file://musl/musl_secure_getenv.patch \
         file://musl/musl_libc_malloc.patch \
         file://musl/musl_gnu_libc-version.patch \
         file://musl/musl_resolver.patch \
-        file://musl/musl_off64_t.patch \
```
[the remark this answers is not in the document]
kraj: this is a generic patch, its not musl specific

recipes-browser/chromium/chromium-gn.inc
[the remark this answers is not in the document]
kraj: done

recipes-browser/chromium/files/64bit_time_t.patch
kraj: submitted an issue with patch
https://bugs.chromium.org/p/chromium/issues/detail?id=1033635

recipes-browser/chromium/files/delete_clang10_warnings.patch
```diff
@@ -0,0 +1,31 @@
+Delete compiler options not available in clang < 10
```
rakuco: Since the flags that are supported change across compiler versions, isn't it better to just add `-Wno-unknown-warning-option`?

kraj: thats already added by meta-clang, this patch fixes the warnings which are now thrown on every file. Slows down the compile and log.do_compile is huge,

recipes-browser/chromium/files/64bit_time_t.patch
```diff
@@ -0,0 +1,54 @@
+Fix build on 32bit arches with 64bit time_t
```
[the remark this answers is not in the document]
kraj: sure yes future versions of glibc too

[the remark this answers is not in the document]
kraj: done

recipes-browser/chromium/files/delete_clang10_warnings.patch
rakuco: Do you mean that despite `-Wno-unknown-warning-option` being passed clang is still complaining about the flags below being passed?

kraj: Well no meta-clang does reduce it from error to warning, `-Wno-error=unknown-warning-option` perhaps using `-Wno-unknown-warning-option` is better here. Since this is always gonna be case

recipes-browser/chromium/files/delete_clang10_warnings.patch
```diff
@@ -0,0 +1,19 @@
...
+Signed-off-by: Khem Raj <raj.khem@gmail.com>
+
+--- a/build/config/compiler/BUILD.gn
++++ b/build/config/compiler/BUILD.gn
+@@ -1481,6 +1481,11 @@ config("default_warnings") {
+ 
+           # TODO(https://crbug.com/999886): Clean up, enable.
+           "-Wno-final-dtor-non-final-class",
++
++	  # Disable unknown warnings, since system clang may not have
```
rakuco: Can you fix the indentation here before landing?

recipes-browser/chromium/files/delete_clang10_warnings.patch
```diff
@@ -0,0 +1,19 @@
+Delete compiler options not available in clang < 10
```
rakuco: I think this needs to be rephrased, as this is no longer specific to clang < 10, but to any version we might be using, since they're usually behind what Chromium expects.

recipes-browser/chromium/files/delete_clang10_warnings.patch
```diff
@@ -0,0 +1,19 @@
...
+
+Upstream-Status: Inappropriate [ Clang/master already supports them]
```
rakuco: I think the explanation here should be _Inappropriate [ upstream expects all these flags to be supported ]_

recipes-browser/chromium/chromium-gn.inc
```diff
@@ -16,17 +16,17 @@ SRC_URI += " \
         file://wrapper-extra-flags.patch \
         file://do-not-specify-march-on-arm.patch \
         file://add_internal_define_armv7ve.patch \
+        file://0001-Fix-building-with-pulseaudio-13.patch \
+        file://delete_clang10_warnings.patch \
```
rakuco: Since this is no longer about clang 10, can you rename the patch?

kraj:
@rakuco all seems ok, can you merge it ?

PR #340, opened 2019-12-13
firefox: Fix buid error of WebRTC

ashie:
WebRTC's code in Firefox isn't buildable against recent Linux kernel.
In addition, it can't build on some platform even if it's old kernel due to the recipe's problem.
This PR fixes these issues.

I confirmed it on zeus, not yet on master.
Because it has unresolved dependency in meta-oe (nodejs-native -> c-ares-native).

### DIFF
--- a/dynamic-layers/rust-layer/recipes-browser/firefox/firefox_68.0esr.bb
+++ b/dynamic-layers/rust-layer/recipes-browser/firefox/firefox_68.0esr.bb
@@ -66,7 +66,7 @@ MOZ_APP_BASE_VERSION = "${@'${PV}'.replace('esr', '')}"
 inherit mozilla rust-common
 
 TOOLCHAIN_pn-firefox = "clang"
-AS_pn-firefox = "${CC}"
+AS = "${CC}"
 
 DISABLE_STATIC=""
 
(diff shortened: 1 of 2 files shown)


ashie:
> Because it has unresolved dependency in meta-oe (nodejs-native -> c-ares-native).

Oops, sorry it's easy mistake.
I forgot to add a new dependency meta-openembedded/meta-networking.

dynamic-layers/rust-layer/recipes-browser/firefox/firefox_68.0esr.bb
```diff
@@ -66,7 +66,7 @@ MOZ_APP_BASE_VERSION = "${@'${PV}'.replace('esr', '')}"
 inherit mozilla rust-common
 
 TOOLCHAIN_pn-firefox = "clang"
-AS_pn-firefox = "${TARGET_PREFIX}clang"
+AS_pn-firefox = "${CC}"
```
kraj: `pn-firefox` seems redundant here. Just setting `AS = "${CC}"` might be enough

ashie: I've removed it, and I confirmed that it's buildable.

ashie: BTW `pn-firefox` for `TOOLCHAIN` seems to be necessary.
I got build error when I remove it due to detecting wrong toolchain.

ashie:
Now I've confirmed also on master.

PR #342, opened 2019-12-17
chromium: Fix build on 32bit arches with 64bit time_t

kraj:
64bit time_t is being added to glibc/musl and will be enabled by default
for 32bit arches atleast its is on musl. This patch ensures that we use
portable defines in a backward compatible way

### DIFF
--- a/recipes-browser/chromium/chromium-gn.inc
+++ b/recipes-browser/chromium/chromium-gn.inc
@@ -20,6 +20,7 @@ SRC_URI += " \
         file://do-not-specify-march-on-arm.patch \
         file://add_internal_define_armv7ve.patch \
         file://delete_not_yet_released_clang_warnings.patch \
+        file://64bit_time_t.patch \
 "
 
 SRC_URI_append_libc-musl = "\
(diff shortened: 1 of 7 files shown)


kraj:
Submitted to chromium gerrit as well

https://chromium-review.googlesource.com/c/chromium/src/+/1972680

rakuco:
> Submitted to chromium gerrit as well

Thanks. I've submitted feedback in the bug report and asked for review in Gerrit (your CL was still in the "WIP" state so no emails had been sent yet).

rakuco:
Given https://chromium-review.googlesource.com/c/chromium/src/+/1972680 was merged upstream, can you backport that commit (following our naming conventions etc, minus the change to `AUTHORS` if it doesn't merge cleanly)?

I also see that you pushed some clang10 fixes to this PR in a separate commit. I suggest creating a separate PR for those changes since they introduce quite a few new patches, some of which have no indication of whether they were sent upstream.

kraj:
I would close this pull since it’s upstream one of these regular updates will bring it in anyway and I have it in my fork for my needed

rakuco:
OK, feel free to then :-)

PR #348, opened 2020-01-07
chromium: rename distro_features_check to features_check

texierp:
Avoid warning due to the class rename in OE-Core.

Fixes:

WARNING: distro_features_check.bbclass is deprecated, please use features_check.bbclass instead

Signed-off-by: Pierre-Jean Texier <pjtexier@koncepto.io>

### DIFF
--- a/recipes-browser/chromium/chromium-gn.inc
+++ b/recipes-browser/chromium/chromium-gn.inc
@@ -2,7 +2,7 @@ require chromium.inc
 require chromium-unbundle.inc
 require gn-utils.inc
 
-inherit distro_features_check gtk-icon-cache qemu
+inherit features_check gtk-icon-cache qemu
 
 # The actual directory name in out/ is irrelevant for GN.
 OUTPUT_DIR = "out/Release"


recipes-browser/chromium/chromium-gn.inc
```diff
@@ -2,7 +2,7 @@ require chromium.inc
 require chromium-unbundle.inc
 require gn-utils.inc
 
-inherit distro_features_check gtk-icon-cache qemu
+inherit features_check gtk-icon-cache qemu
```
kraj: It will break older releases. We support multiple releases with master back until thud. this will mean that we have to branch out for Zeus and older branches before accepting this change. Given resources we have, it might be extra work to maintain two branches, but I would like to hear from @otavio @rakuco @msisov as well on what they think about it.

otavio: I am fine in maintaining it for one more cycle and branch out. Many people still uses `thud`.

kraj: seems fine. since 3.1 is probably be LTS, we should perhaps we should see if we can drop support for all prior releases until 3.1 then.

otavio: After rethinking, I agree with you.

Supporting 3.1 does make more sense and also reduces the number of builds to do.

otavio:
Everyone is Ok in we merge it now ?

kraj:
> Everyone is Ok in we merge it now ?

have we officially said that we will not support anything below 3.1 ?
if yes then its ok.

otavio:
If we want to reduce our work overload as well as push for people to upgrade to it, it makes sense to me. @msisov @rakuco ?

kraj:
I am totally fine with dropping older releases.

msisov:
This is going to be very inconvenient, tbh. We have a well defined process of testing with stable branches and pointing our clients to use this repository. AFAIK, most of the yocto based projects use stable branches (2.7, 3.0 branches).

otavio:
> This is going to be very inconvenient, tbh. We have a well defined process of testing with stable branches and pointing our clients to use this repository. AFAIK, many of the yocto based projects I know use stable branches (2.7, 3.0 branches).

Eventually, we will need to do that. We can do that now or in the future.

How this is done on your side @msisov currently? and what versions do you support nowadays?

kraj:
> This is going to be very inconvenient, tbh. We have a well defined process of testing with stable branches and pointing our clients to use this repository. AFAIK, many of the yocto based projects I know use stable branches (2.7, 3.0 branches).

since 3.1 will be LTS perhaps we could use that as next stop for mimimal support and
create oldstable or some such branch which supports older releases. It wont be sustainable
to support so many yocto relesses just with one master branch

msisov:
> > This is going to be very inconvenient, tbh. We have a well defined process of testing with stable branches and pointing our clients to use this repository. AFAIK, many of the yocto based projects I know use stable branches (2.7, 3.0 branches).
> 
> Eventually, we will need to do that. We can do that now or in the future.
> 
> How this is done on your side @msisov currently? and what versions do you support nowadays?

Thud and Warrior. We haven't upgraded to Zeus yet. https://gitlab.com/browsers/meta-browser/pipelines

otavio:
And what is the plan @msisov, support it for how long?

msisov:
@otavio well, if you think master must be master, let's merge this. I don't have strong opinion on that. It will just make my life a little bit harder.

otavio:
I am keen to keep it a little longer but not forever. @Domarys please backport the hack made on https://github.com/Igalia/meta-webkit/commit/d2f83d9c824eb913ca713ffec86f8ddeb65adeda so we keep this for some time. 

@msisov this will make your life easier ;-)

shr-project:
It might be easier to work around this by backporting the features_check.bbclass for zeus and older builds, like I did for meta-qt5:
https://github.com/webOS-ports/meta-webos-ports/commit/fb3e94ba4a200eaba989a2871a4a5344e2162278

otavio:
> It might be easier to work around this by backporting the features_check.bbclass for zeus and older builds, like I did for meta-qt5:
> [webOS-ports/meta-webos-ports@fb3e94b](https://github.com/webOS-ports/meta-webos-ports/commit/fb3e94ba4a200eaba989a2871a4a5344e2162278)

You mean @msisov to do it on his side?

shr-project:
> You mean @msisov to do it on his side?

yes

msisov:
ok. let's merge this patch then.

Issue #367, opened 2020-03-24
chromium-ozone-wayland-80.0.3987.132 missing two includes


MastaG:
Hi I'm currently unable to create a PR.

But with the latest master branches of oe-core and meta-clang I need to modify two files in order to successfully build chromium ozone wayland.

third_party/perfetto/include/perfetto/base/task_runner.h: error: ‘uint32_t’ does not name a type
needs #include cstdint

base/trace_event/trace_event_memory_overhead.h:
needs #include string

After these two includes, I can build it 100%.

kraj:
I have a patch for clang10 here https://github.com/YoeDistro/meta-browser/commit/1359b9620c8a26c9e6e4fe5226ef8a6752b702f4 give it a shot

MastaG:
Thanks I've applied all of them except for the last one:

Applying patch 0001-Remove-glslang-pool_allocator-setAllocator.patch
patching file third_party/glslang/src/glslang/Include/PoolAlloc.h
Hunk #1 FAILED at 304.
1 out of 1 hunk FAILED -- rejects in file third_party/glslang/src/glslang/Include/PoolAlloc.h
Patch 0001-Remove-glslang-pool_allocator-setAllocator.patch can be reverse-applied

I think it's already upstream.

Building now, will report back.

MastaG:
all OK :)
Please send a PR, just without the last patch. 

Thanks a lot!

kraj:
@MastaG please go ahead with the pull since you have tested it more recently than me and have uptodate content to pull from.

MastaG:
Will do, thanks!

PR #368, opened 2020-03-25
chromium: fix building with clang 10

MastaG:
Thanks @kraj  for the original patch.

### DIFF
--- a/recipes-browser/chromium/chromium-gn.inc
+++ b/recipes-browser/chromium/chromium-gn.inc
@@ -19,6 +19,10 @@ SRC_URI += " \
         file://do-not-specify-march-on-arm.patch \
         file://add_internal_define_armv7ve.patch \
         file://delete_not_yet_released_clang_warnings.patch \
+        file://std_size_t.patch \
+        file://stdint.patch \
+        file://include_string.patch \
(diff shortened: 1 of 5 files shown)


otavio:
All those looks good to go, I wonder if it did not break old versions.

@rakuco do you mind triggering a build with this?

rakuco:
I'm launching a warrior-intel-corei7-64 to double-check. I still need to check those patches a bit more closely though, at least some of them look like libstdc++ fixes, not clang ones (especially since Chromium itself is always using a bleeding-edge clang version).

MastaG:
Well I can speak for the missing includes that they're definitely required to build ozone-wayland for armv7.
I reported them here: https://github.com/OSSystems/meta-browser/issues/367

The other patches I can't review since I'm no C++ developer.

But I can verify that they do not break the build.

otavio:
@kraj @rakuco ?

kraj:
These patches have been working ok for me for few months

MastaG:
I can only speak for 32bit arm on the OE master branches.
And I've only did a clang with ld-is-ldd build.
So I can't speak for other architectures, older branches or compiling with gcc.

Issue #357, opened 2020-02-21
chromium-x11 soft-crashes frequently with 'Aw, snap'


schnitzeltony:
Don't exactly know when it started but it was there with 79.x already. Have decided to wait for 80.x hoping it might go away but it didn't:

On a fresh image chromium (no extensions) frequently displays:

![chromium-error](https://user-images.githubusercontent.com/2571823/75065563-5fd30700-54e9-11ea-8f73-070676df6974.png)

Reloading helps but the error comes back soon - sometimes by just scrolling a page.

Starting chrome from console I get:

> [3290:7:0221/202546.686655:ERROR:command_buffer_proxy_impl.cc(125)] ContextResult::kTransientFailure: Failed to send GpuChannelMsg_CreateCommandBuffer.
> [3290:1:0221/202547.114520:ERROR:child_process_sandbox_support_impl_linux.cc(79)] FontService unique font name matching request did not receive a response.
> [3290:1:0221/202547.115298:ERROR:child_process_sandbox_support_impl_linux.cc(79)] FontService unique font name matching request did not receive a response.
> [3361:1:0221/202606.303449:ERROR:child_process_sandbox_support_impl_linux.cc(79)] FontService unique font name matching request did not receive a response.
> [3361:1:0221/202606.306028:ERROR:child_process_sandbox_support_impl_linux.cc(79)] FontService unique font name matching request did not receive a response.
> [3361:1:0221/202606.309189:ERROR:child_process_sandbox_support_impl_linux.cc(79)] FontService unique font name matching request did not receive a response.
> [3361:1:0221/202606.314838:ERROR:child_process_sandbox_support_impl_linux.cc(79)] FontService unique font name matching request did not receive a response.
> ../../sandbox/linux/seccomp-bpf-helpers/sigsys_handlers.cc:**CRASHING**:seccomp-bpf failure in syscall 0407

Any ideas?

kraj:
this perhaps is related to pread64 syscall or some other not being addressed by libseccomp
is it possible for you to try the same build with glibc 2.30 ( zeus ) and see if it still happens ?

schnitzeltony:
@kraj My machine is weak and building 24/7 for other issues. Same answer I sent to you for other issue (was it yesterday?): Will try this weekend - have an extra free monday :grin:

schnitzeltony:
Some common comments
1. @kraj I don't have the resources to go back to glibc-2.30 - but I think you are right
2. Nobody else?

[...]

PR #371, opened 2020-04-22
chromium: Fix frequent 'Aw, snap'on 32bit systems with glibc >= 2.31

schnitzeltony:
* All necessary background infomation is in patch added
* Fixes [1]

[1] https://github.com/OSSystems/meta-browser/issues/357


otavio:
@kraj @schnitzeltony could someone send this upstream? seems something useful.

Issue #355, opened 2020-02-14
Firefox fails to build on aarch64, zeus branch


MikeLooijmans:
With "zeus" branch of OE, and today's "master" of meta-browser and meta-rust, the build fails with the following message. It did compile okay on warrior a few weeks ago.

|  9:24.86 media/webrtc/trunk/webrtc/api/audio_codecs/ilbc/audio_decoder_ilbc_gn
|  9:26.03 error[E0506]: cannot assign to `self.input.cached_token` because it is borrowed
|  9:26.03    --> /usr/src/debug/firefox/68.0esr-r0/firefox-68.0/third_party/rust/cssparser/src/parser.rs:584:17
|  9:26.03     |
|  9:26.03 559 |     pub fn next_including_whitespace_and_comments(&mut self) -> Result<&Token<'i>, BasicParseError<'i>> {
|  9:26.03     |                                                   - let's call the lifetime of this reference `'1`
|  9:26.03 ...
|  9:26.03 572 |             Some(ref cached_token)
|  9:26.03     |                  ---------------- borrow of `self.input.cached_token` occurs here
|  9:26.03 ...
|  9:26.03 584 |                 self.input.cached_token = Some(CachedToken {
|  9:26.03     |                 ^^^^^^^^^^^^^^^^^^^^^^^ assignment to borrowed `self.input.cached_token` occurs here
|  9:26.03 ...
|  9:26.03 596 |         Ok(token)
|  9:26.03     |         --------- returning this value requires that `self.input.cached_token.0` is borrowed for `'1`
|  9:26.15 media/webrtc/trunk/webrtc/api/audio_codecs/ilbc/audio_encoder_ilbc_gn
|  9:26.49 error: aborting due to previous error
|  9:26.49 For more information about this error, try `rustc --explain E0506`.
|  9:26.51 error: could not compile `cssparser`.

cosmo0920:
I'd also hit the issue.

Using Rust 1.37.0 can be avoided compilation failure.

I'm using the following settings in local.conf:

```
PREFERRED_VERSION_rust-native ?= "1.37.0"
PREFERRED_VERSION_rust-cross-aarch64 ?= "1.37.0"
PREFERRED_VERSION_rust-llvm-native ?= "1.37.0"
PREFERRED_VERSION_libstd-rs ?= "1.37.0"
PREFERRED_VERSION_cargo-native ?= "1.37.0"

```

MastaG:
Thanks!

Issue #338, opened 2019-12-09
Firefox fails to build for armv7


MastaG:
Hi I'm a bit new to firefox/rust, but it fails to build for armv7.

```

------
0:11.06    Compiling webrender v0.60.0 (/home/mastag/src/odroid-oe-core/build/tmp/work/cortexa15t2hf-neon-vfpv4-oe-linux-gnueabi/firefox/68.0esr-r0/firefox-68.0/gfx/wr/webrender)
 0:11.08 error: failed to run custom build command for `style v0.0.1 (/home/mastag/src/odroid-oe-core/build/tmp/work/cortexa15t2hf-neon-vfpv4-oe-linux-gnueabi/firefox/68.0esr-r0/firefox-68.0/servo/components/style)`
 0:11.08 Caused by:
 0:11.08   process didn't exit successfully: `/home/mastag/src/odroid-oe-core/build/tmp/work/cortexa15t2hf-neon-vfpv4-oe-linux-gnueabi/firefox/68.0esr-r0/firefox-68.0/firefox-build-dir/release/build/style-01cf4f6b28ee30c3/build-script-build` (exit code: 1)
------
 0:11.08 --- stderr
 0:11.08 Could not find platform independent libraries <prefix>
 0:11.08 Could not find platform dependent libraries <exec_prefix>
 0:11.08 Consider setting $PYTHONHOME to <prefix>[:<exec_prefix>]
 0:11.08 ImportError: No module named site
 0:11.08 warning: build failed, waiting for other jobs to finish..
------
 4:43.96 error: build failed
 4:43.96 make[4]: *** [/home/mastag/src/odroid-oe-core/build/tmp/work/cortexa15t2hf-neon-vfpv4-oe-linux-gnueabi/firefox/68.0esr-r0/firefox-68.0/config/makefiles/rust.mk:240: force-cargo-library-build] Error 101
 4:43.96 make[3]: *** [/home/mastag/src/odroid-oe-core/build/tmp/work/cortexa15t2hf-neon-vfpv4-oe-linux-gnueabi/firefox/68.0esr-r0/firefox-68.0/config/recurse.mk:74: toolkit/library/rust/target] Error 2
 4:43.96 make[3]: *** Waiting for unfinished jobs....
[...]

PR #378, opened 2020-05-28
Firefox changes to make it compatible with rust-1.39 instead of even older rust-1.37

shr-project:
I'll try to get new zeus branch in meta-rust which will include 1.39 version + few build fixes, it will be better than using arbitrary commit in meta-rust (like we need to use now to get 1.37 version which is latest compatible with current firefox).

I've WIP patch for firefox upgrade to 68.9.0esr release candidate, which includes few more patches, will send it after the final version is released.

Fixes https://github.com/OSSystems/meta-browser/issues/355


shr-project:
I haven't tested it in runtime, I am interested only in fixing the build, because I've unfortunately used it in https://github.com/shr-project/test-oe-build-time

kraj:
@ashie can you also take a look please

shr-project:
To build this with latest meta-rust you need to select older rust, with:
```
RUST_VERSION = "1.39.0"
PREFERRED_VERSION_rust-native ?= "${RUST_VERSION}"
PREFERRED_VERSION_rust-cross-aarch64 ?= "${RUST_VERSION}"
PREFERRED_VERSION_rust-llvm-native ?= "${RUST_VERSION}"
PREFERRED_VERSION_libstd-rs ?= "${RUST_VERSION}"
PREFERRED_VERSION_cargo-native ?= "${RUST_VERSION}"
```

shr-project:
Tested with zeus and with dunfell where it also needs python and python2.7 in HOSTTOLS (as in https://github.com/shr-project/meta-browser/commit/c18e150731e2875b3c523eaa46a9662267de2c27 )

ashie:
> @ashie can you also take a look please

I see.

dynamic-layers/rust-layer/recipes-browser/firefox/firefox_68.0esr.bb
```diff
@@ -175,3 +175,5 @@ PRIVATE_LIBS += " \
     libssl3.so \
     libsoftokn3.so \
 "
+
+CLEANBROKEN = "1"
```
ashie, line 177: >    raise Exception('Could not detect environment shell!')

Could you try to set correct shell like this?

https://github.com/OSSystems/meta-browser/blob/5e58ff65facc27a3054a4d3f97d3329342d49afa/classes/mozilla.bbclass#L64
https://github.com/OSSystems/meta-browser/blob/5e58ff65facc27a3054a4d3f97d3329342d49afa/classes/mozilla.bbclass#L30

shr-project:
68.9.0ESR release candidate was released today as it was, so I've added it to PR

dynamic-layers/rust-layer/recipes-browser/firefox/firefox_68.0esr.bb
shr-project: Then it fails, because mozconfig doesn't exist in ${B}

```
/OE/build/oe-core/workspace/sources/firefox$ export SHELL=/bin/sh
/OE/build/oe-core/workspace/sources/firefox$ make clean
./mach clobber
MOZCONFIG environment variable refers to a path that does not exist: /OE/build/oe-core/workspace/sources/firefox/mozconfig
make: *** [GNUmakefile:12: clean] Error 1
```

ashie: Memo for me: what does `CLEANBROKEN` do:

* https://www.yoctoproject.org/docs/latest/ref-manual/ref-manual.html#var-CLEANBROKEN
* https://www.yoctoproject.org/docs/latest/ref-manual/ref-manual.html#migration-1.8-rebuild-improvements
* https://git.congatec.com/yocto/poky/blob/fido/meta/classes/autotools.bbclass#L113

ashie: > MOZCONFIG environment variable refers to a path that does not exist: /OE/build/oe-core/workspace/sources/firefox/mozconfig
make: *** [GNUmakefile:12: clean] Error 1

This file is created at here:

https://github.com/OSSystems/meta-browser/blob/5e58ff65facc27a3054a4d3f97d3329342d49afa/classes/mozilla.bbclass#L45

I think we can resolve it by putting the file to `.mozconfig` (it's the default path of mozconfig) instead of specifying the path by `MOZCONFIG`.
`./mach clobber` doesn't fail even if `.mozconfig` doesn't exist.

dynamic-layers/rust-layer/recipes-browser/firefox/firefox/debian-hacks/Add-a-2-minutes-timeout-on-xpcshell-tests.patch
```diff
@@ -1,13 +1,14 @@
+From 9bbdf6b9de2049d072a70644de552b5d0929065b Mon Sep 17 00:00:00 2001
 From: Mike Hommey <mh@glandium.org>
 Date: Mon, 27 Dec 2010 10:44:28 +0100
-Subject: Add a 2 minutes timeout on xpcshell tests
+Subject: [PATCH] Add a 2 minutes timeout on xpcshell tests
```
ashie, line 4: Is there any reason that you change the format of debian patches?
Here is the original one:

  * https://salsa.debian.org/mozilla-team/firefox/-/blob/debian/68.8.0esr-1/debian/patches/debian-hacks/Add-a-2-minutes-timeout-on-xpcshell-tests.patch

I prefer that we don't change it to suppress meaningless differentials.

dynamic-layers/rust-layer/recipes-browser/firefox/firefox_68.0esr.bb
ashie: ... but I'm not necessarily expecting to fix it now.
Please let me know if it's hard to fix.

dynamic-layers/rust-layer/recipes-browser/firefox/firefox/debian-hacks/Add-a-2-minutes-timeout-on-xpcshell-tests.patch
shr-project: Just because devtool updates it this way, I also had to fix devtool first to work correctly with firefox :/ because of patching files listed in .gitignore and not having proper patch headers, see:
https://lists.openembedded.org/g/openembedded-core/message/138923?p=,,,20,0,0,0::Created,,devtool,20,2,0,74554370
https://lists.openembedded.org/g/openembedded-core/message/138928?p=,,,20,0,0,0::Created,,devtool,20,2,0,74554376

dynamic-layers/rust-layer/recipes-browser/firefox/firefox_68.0esr.bb
shr-project: I don't know how hard is to fix, but CLEANBROKEN is simple work around, so I suggest to fix it separately later (preferably by someone else in separate PR :))

I don't use firefox at all, adding CLEANBROKEN was needed to be able to use devtool (clean task was failing during "devtool finish").

shr-project:
https://ftp.mozilla.org/pub/firefox/releases/68.9.0esr/source/ is there again, I'll check if it's the same as https://ftp.mozilla.org/pub/firefox/candidates/68.9.0esr-candidates/build1/source/ I'm using now and add the upgrade (so we can drop 4 more patches).

shr-project:
68.9.0 upgraded added, there is still that python issue with dunfell as discussed in https://github.com/OSSystems/meta-browser/issues/338 - IMHO more important than CLEANBROKEN work around.

I've tested it only with zeus and with dunfell (where I've added python, python2.7 to HOSTTOOLS).

PR #385, opened 2020-06-12
chromium: Add missing include to fix build in oe-master environment

schnitzeltony:


schnitzeltony:
Done

rakuco:
Thanks for the patch. For the record, I've sent #387 which just replaces this patch with the upstream version of the same fix for better accountability.

PR #395, opened 2020-06-26
chromium: Add a unified ozone wayland/x11/headless recipe for Chromium

caneraltinbasak:
Added a new chromium-ozone-headless recipe for building Chromium for
headless systems. Chromium will output to a png file pointed by
--ozone-dump-file. Refer to Chromium ozone documents for further
information.


msisov:
I don't think we need a separate recipe for that. The goal of the Ozone platform is to make it possible to run a single binary with multiple backends.

Thus, I'd suggest you to just rename the existing ozone/wayland recipe to chromium-ozone and use headless from there (it's actually enabled). https://github.com/OSSystems/meta-browser/blob/master/recipes-browser/chromium/chromium-ozone-wayland_83.0.4103.116.bb#L16

What is more, we are in a process of getting rid of non-Ozone X11 linux config in upstream and making Ozone as a feature flag. Thus, the chromium-x11 recipe won't be needed at all (in the future). https://bugs.chromium.org/p/chromium/issues/detail?id=1085700

msisov:
You can specify what platform you want to use by running chromium with --ozone-platform=headless/wayland/x11 (if all these platforms in question are built)

caneraltinbasak:
> Thus, I'd suggest you to just rename the existing ozone/wayland recipe to chromium-ozone and use headless from there (it's actually enabled). https://github.com/OSSystems/meta-browser/blob/master/recipes-browser/chromium/chromium-ozone-wayland_83.0.4103.116.bb#L16

It still requires wayland in DISTRO_FEATURES, and has different dependencies than headless.  It won't compile for non-graphical distros.

caneraltinbasak:
I'll rename ozone-wayland to ozone. I'll make DEPENDS and build flags conditional with bb.utils.contains('DISTRO_FEATURES'). Then x11 and wayland can be enabled if they are in DISTRO_FEATURES.

PR #399, opened 2020-07-09
firefox: Provide symbolic link for python2.7 when firefox 68esr is built

lmajewski:
This is a hack for allow building the firefox 68esr on Centos/Fedora.
Detailed description of the problem:
https://github.com/OSSystems/meta-browser/issues/338

Signed-off-by: Lukasz Majewski <lukma@denx.de>

### DIFF
--- a/meta-firefox/classes/mozilla.bbclass
+++ b/meta-firefox/classes/mozilla.bbclass
@@ -3,7 +3,7 @@ DEPENDS += "gnu-config-native virtual/libintl libxt libxi zip-native gtk+"
 
 SRC_URI += "file://mozconfig"
 
-inherit pkgconfig
+inherit pkgconfig pythonnative
 
 EXTRA_OECONF = "--target=${TARGET_SYS} --host=${BUILD_SYS} \
                 --with-toolchain-prefix=${TARGET_SYS}- \
@@ -62,6 +62,15 @@ mozilla_do_configure() {
 }
 
 mozilla_do_compile() {
+
+	# This is a hack/workaround necessary for building firefox 68esr on
+	# Centos 7.6/Fedora
+	# It doesn't do any harm on other distros - like Ubuntu 18.04 or Debian 10
+	# Moreover, it mimics 'include pythonnative', which cannot be used as it
+	# causes errors during firefox 68esr configuration stage
+	ln -snf python-native/python2.7 ${STAGING_BINDIR_NATIVE}/python2.7
+	ln -snf python-native/python2.7 ${STAGING_BINDIR_NATIVE}/python
+
 	mozilla_run_mach build
 }
 
--- a/meta-firefox/conf/layer.conf
+++ b/meta-firefox/conf/layer.conf
@@ -4,14 +4,6 @@ BBPATH .= ":${LAYERDIR}"
 # We have a recipes directory, add to BBFILES
 BBFILES += "${LAYERDIR}/recipes-*/*/*.bb ${LAYERDIR}/recipes-*/*/*.bbappend"
 
-# this is only temporary to allow building firefox on hosts without any python2
-# were it would fail like this:
-# ERROR: firefox-68.9.0esr-r0 do_configure: Execution of '/OE/build/oe-core/tmp-glibc/work/core2-64-oe-linux/firefox/68.9.0esr-r0/temp/run.do_configure.2697230' failed with exit code 127:
-# ./mach: 9: exec: python: not found
-# this should be removed when firefox is upgraded to 78 ESR or newer which contains:
-# https://bugzilla.mozilla.org/show_bug.cgi?id=1473498
-HOSTTOOLS += "python python2.7"
-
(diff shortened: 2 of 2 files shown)


classes/mozilla.bbclass
```diff
@@ -62,6 +62,16 @@ mozilla_do_configure() {
...
 
 mozilla_do_compile() {
+
+# This is a hack/workaround necessary for building firefox 68esr on
+# Centos 7.6/Fedora
+# It shall not do any harm on other distros - like Ubuntu 18.04 or Debian 10
+# It mimics 'include pythonnative', which causes errors during firefox 68esr
+# configure stage
+	cd ${STAGING_BINDIR_NATIVE}
+	ln -sf ./python-native/python2.7 python2.7
```
shr-project, line 72: ```ln -snf python-native/python2.7 ${STAGING_BINDIR_NATIVE}/python2.7```

Does the same job without both cd calls, doesn't it?

otavio: Also, the comment should be aligned with the code.

ashie: I don't understand how does it resolve or avoid https://github.com/OSSystems/meta-browser/issues/338#issuecomment-653534486.
At least `python-native` is required in `DEPENDS` to use it.

lmajewski: @ashie Python-native is already installed to recipe-sysroot-native.

I've also filled in the bug report at bugzilla for firefox project - the reply there also shed some light to the problem:
https://bugzilla.mozilla.org/show_bug.cgi?id=1651641

lmajewski: IMHO - the best solution would be to use include pythonnative, which would be the same (and consistent) version for every Yocto/OE release.

However, this causes some issues with firefox -c configure (the error with _virtualenv setup). 
My observation is that for firefox -c configure the HOST python2.7 is used, then for firefox -c compile, the style component requires newer python - which is in recipe-sysroot-native.

otavio: In this case, @ashie is right, as we need to make the dependency explicit. Are you testing this on `master` or `dunfell`? I think in this case, `python2` would not be available by default.

lmajewski: I'm testing it with Thud (2.6) YPRR on Centos 7.6, Debian10, Ubuntu 18.04. I did not have a chance to run it on newer Yocto/OE release.

classes/mozilla.bbclass
```diff
@@ -62,6 +62,14 @@ mozilla_do_configure() {
 }
 
 mozilla_do_compile() {
+
+	# This is a hack/workaround necessary for building firefox 68esr on
+	# Centos 7.6/Fedora
+	# It doesn't do any harm on other distros - like Ubuntu 18.04 or Debian 10
+	# Moreover, it mimics 'include pythonnative', which cannot be used as it
+	# causes errors during firefox 68esr configuration stage
+	ln -snf python-native/python2.7 ${STAGING_BINDIR_NATIVE}/python2.7
```
kraj, line 71: instead of symlink does setting `PYTHON="${STAGING_BINDIR_NATIVE}/python-native/python2.7"` help ?

lmajewski: Unfortunately, this is not working 
`|  7:14.04 --- stderr`
`|  7:14.04 Could not find platform independent libraries <prefix>`
`|  7:14.04 Could not find platform dependent libraries <exec_prefix>`
`|  7:14.04 Consider setting $PYTHONHOME to <prefix>[:<exec_prefix>]`
`|  7:14.04 ImportError: No module named site`
`|  7:14.04 warning: build failed, waiting for other jobs to finish...`
`|  7:14.04 modules/fdlibm/src`

Here the problem is that cargo doesn't work correctly with PYTHON variable passed to it - as described in fix for the following bug report:
https://bugzilla.mozilla.org/show_bug.cgi?id=1651641

lmajewski: Do we have any agreement on this?
If this change is not acceptable to `-master` I can prepare a PR for Thud YPRR (2.6) on which I do work.

otavio: It seems upstream did a change which we might use:

https://hg.mozilla.org/integration/autoland/rev/cbfe3335cfe5

Could it be backported?

otavio (approved):

shr-project:
@otavio I think this doesn't work (it definitely doesn't build for me, I thought I have added negative review here, but now I don't see it anywhere :/.

otavio:
Oh! I assumed from the code you had "fixed it". You did push it few days ago and seems fine.

Are you going to prepare a revert? Sorry for the confusion.

shr-project:
> Are you going to prepare a revert? Sorry for the confusion.

Done. No problem it was my fault for not submitting the comment I wrote or writing it in wrong PR or something. I've merged the revert and includes similar reasoning as what I wrote somewhere before :).

PR #413, opened 2020-09-29
chromium: add patch to include uint64_t in sim_hash namespace

viatsk:
Previously failing to compile sim_hash.h with error
| In file included from ../../components/federated_learning/sim_hash.cc:5:
| ../../components/federated_learning/sim_hash.h:20:15: error: unknown type name 'uint64_t'
|   void SetBit(uint64_t pos);

Patch 0001-Fix-lack-of-uint64_t-in-sim_hash-namespace.patch fixes
this error, but should be removed as soon as the chromium version
is updated to 86.x.xxxx.xx+.

### DIFF
--- a/recipes-browser/chromium/chromium-gn.inc
+++ b/recipes-browser/chromium/chromium-gn.inc
@@ -21,6 +21,7 @@ SRC_URI += " \
         file://0001-IWYU-missing-include-for-std-vector-usage-in-ozone-p.patch \
         file://0001-Revert-ui-gfx-linux-Remove-2-unnecessary-preprocesso.patch \
         file://0001-Turn-TouchMove-to-async-after-the-first-GestureScrol.patch \
+        file://0001-IWYU-add-includes-for-uint64_t-and-std-string.patch \
 "
 
 SRC_URI_append_libc-musl = "\
--- a/recipes-browser/chromium/files/0001-IWYU-add-includes-for-uint64_t-and-std-string.patch
+++ b/recipes-browser/chromium/files/0001-IWYU-add-includes-for-uint64_t-and-std-string.patch
@@ -1,0 +1,37 @@
+Upstream-Status: Backport
+
+Signed-off-by: Stacy Gaikovaia <stacy.gaikovaia@windriver.com>
+---
+From af06ca9d46d46302c100f07e499d8aaa9ab32364^! Tue Jan 30 15:58:38 2020
+Author:     Stephan Hartmann <stha09@googlemail.com>
+AuthorDate: 2020-06-30 15:58:38 +0000
+Commit:     Commit Bot <commit-bot@chromium.org>
+CommitDate: 2020-06-30 15:58:38 +0000
+
+IWYU: add includes for uint64_t and std::string
+
+
+Bug: 819294
+Change-Id: Iae844f8b6cf6bba879cb320c95ebf88f65e1385f
+Reviewed-on: https://chromium-review.googlesource.com/c/chromium/src/+/2273161
+Reviewed-by: Yao Xiao <yaoxia@chromium.org>
+Commit-Queue: Yao Xiao <yaoxia@chromium.org>
+Cr-Commit-Position: refs/heads/master@{#783978}
+---
+ components/federated_learning/sim_hash.h | 2 ++
+ 1 file changed, 2 insertions(+)
+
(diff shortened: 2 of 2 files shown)


otavio:
Please rebase the branch and send it again.

kraj:
once rebased the patch looks fine to me.

recipes-browser/chromium/files/0001-Fix-lack-of-uint64_t-in-sim_hash-namespace.patch
```diff
@@ -0,0 +1,65 @@
+From 6284aee59f251bfe4b52d200b530ae8a98186adb Mon Sep 17 00:00:00 2001
```
msisov, line 1: when backporting patches, please follow the style guide. In this case -

Upstream-Status: Backport

Backported from - {link}.

Signed-off-by: Name {name@mail.com}

msisov (commented):
What version of clang do you use that you was able to hit this error?

recipes-browser/chromium/files/0001-Fix-lack-of-uint64_t-in-sim_hash-namespace.patch
msisov: If you have doubts how to properly style it, check other patches we have.

viatsk: Is it still considered a backport if I am the author of this patch (not chromium upstream)?

kraj: yes, backport means its cherry-picked from upstream, its unrelated to author.

rakuco: To clarify: the idea is to backport the upstream fix (https://chromium-review.googlesource.com/c/chromium/src/+/2273161). You sign-off-by above the change, but maintain the proper authorship in the commit message.

viatsk: @rakuco Sorry, to be clear, I didn't look at the change in chromium at all. Instead, I made a different but similar set of changes based on information I've read here: https://stackoverflow.com/questions/13642827/cstdint-vs-stdint-h

The difference is that my patch includes also pre-pending uses of uint64_t in the namespace with std::. Do I still have to cite that upstream chromium patch if I didn't use it for reference, and my patch does extra things?

kraj: Upstream fix is always preferred, since that means we can easily drop it when we move past a new release which includes it. I would suggest to replace the proposed with the patch that got committed upstream. if you think your patch should be included upstream I would suggest to propose it to chromium gerrit

otavio (approved):

recipes-browser/chromium/files/0001-Fix-lack-of-uint64_t-in-sim_hash-namespace.patch
viatsk: Ok, thank you for all the feedback. I have changed my PR to include the backport only.

viatsk:
> What version of clang do you use that you was able to hit this error?

It was clang-11.

recipes-browser/chromium/files/0001-Fix-lack-of-uint64_t-in-sim_hash-namespace.patch
rakuco: Thanks!

viatsk:
Why was this patch removed? My copy of the browser is once again failing to build with clang-11, with the same error as before this patch was submitted.

kraj:
> Why was this patch removed? My copy of the browser is once again failing to build with clang-11, with the same error as before this patch was submitted.

we are at 86.0.4240.198 and the original comment said we wont need it with 86.0.4240.198, is that not the case ?

viatsk:
> we are at 86.0.4240.198 and the original comment said we wont need it with 86.0.4240.198, is that not the case ?

Yes, it's my mistake - I accidentally had linked my build against the older source code, so it was actually building the old version, sorry.

PR #426, opened 2020-12-17
chromium: Add mime-xdg to get rid of build warning

viatsk:
When building chromium without including mime-xdg, bitbake throws the
following warning:

WARNING: chromium-x11-86.0.4240.111-r0 do_package_qa:
QA Issue: package contains desktop file with key 'MimeType' but does not inhert mime

In order to deal with this warning, have the chromium recipe inherit
the mime-xdg recipe.

Signed-off-by: Stacy Gaikovaia <stacygaikovaia@gmail.com>


rakuco (approved):
lgtm, could you just keep the list of classes being `inherit`ed sorted?

viatsk:
I've updated the PR :)

Issue #431, opened 2021-01-18
Missing libatomic.so.1.2.0 (by native clang) in Centos8


yifan19:
## Issue:
There is an implicit dependency for clang to `libatomic1` when trying to build `yocto_native/brotli`

## Setup:
Windriver linux building chromium-x11

also reproduced in poky with everything default
with the following change
```
config/local.conf
++ IMAGE_INSTALL_append = " chromium-x11" 
```

and adding all required layers (python2, open-embedded and clang)

`bitbake core-image-x11`

## Error Msg:
ld.lld: error cannot open /usr/lib64/libatomic.so.1.2.0: No such file or directory

## Version:
- chromium-x11 86.0.4240.198-r0  (meta-browser at 1aef1b7)
- Centos 8
- everything else pulled from master

yifan19:
Might be able to revert this if we link against llvm's run-time libraries
Running a test on Centos8

PR #436, opened 2021-01-22

hongxu-jia:
Since chromium use clang as toolchain and clang delegates atomics to
runtime library instead of builtins, Link with latomic in TARGET_LDFLAGS
rather than in BUILD.gn, to avoid build failure due to no libatomic on
host (such as CentOS 8)

### DIFF
--- a/recipes-browser/chromium/chromium-x11_87.0.4280.141.bb
+++ b/recipes-browser/chromium/chromium-x11_87.0.4280.141.bb
@@ -2,6 +2,10 @@ require chromium-gn.inc
 
 REQUIRED_DISTRO_FEATURES = "x11"
 
+SRC_URI += " \
+    file://0001-remove-link-to-libatomic-from-BUILD.gn.patch \
+"
+
 DEPENDS += "\
         libx11 \
         libxcomposite \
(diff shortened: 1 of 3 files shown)


yifan19:
relates to this issue 
https://github.com/OSSystems/meta-browser/issues/431

recipes-browser/chromium/chromium-x11_87.0.4280.141.bb
```diff
@@ -2,6 +2,10 @@ require chromium-gn.inc
 
 REQUIRED_DISTRO_FEATURES = "x11"
 
+SRC_URI += " \
```
msisov, line 5: This doesn't look like chromium-x11 only patch. Please move that to chromium-gn.inc

recipes-browser/chromium/chromium-x11_87.0.4280.141.bb
```diff
@@ -27,3 +31,9 @@ GN_ARGS += "use_ozone=false"
 # chromium-ozone-wayland recipes, and the former used to be called just
 # "chromium".
 PROVIDES = "chromium"
+
+# Since chromium use clang as toolchain and clang delegates atomics to
```
msisov, line 35: s/Since chromium use/Since chromium uses/

recipes-browser/chromium/files/0001-remove-link-to-libatomic-from-BUILD.gn.patch
```diff
@@ -0,0 +1,60 @@
+From 01c899455bff7ab0322542c832d4f7c3359e9139 Mon Sep 17 00:00:00 2001
```
msisov, line 1: Please use correct format for patches. See other files in this (files) directory.

hongxu-jia: OK, I will resubmit pull request

msisov: you could also push into the same branch from which your new PR comes from.

PR #437, opened 2021-01-22
chromium: add patch that enables VA-API with ozone/wayland.

msisov:
Not only the HWs were tested. The patch was mostly tested
with Intel Gen8 and Gen9. Also, with AMD RX460. So, bear that
in mind when you enable va-api.

In order to have it running, one must enable VA-API and
proprietary codecs in Chromium
PACKAGECONFIG_pn-chromium-ozone-wayland_append = " proprietary-codecs
use_vaapi".

### DIFF
--- a/recipes-browser/chromium/chromium-ozone-wayland_87.0.4280.141.bb
+++ b/recipes-browser/chromium/chromium-ozone-wayland_87.0.4280.141.bb
@@ -10,6 +10,14 @@ DEPENDS += "\
         wayland-native \
 "
 
+SRC_URI += "\
+        file://0001-ozone-add-va-api-support-to-wayland.patch \
+"
+
+# Enables VA-API for Ozone/Wayland. Remember to also use proprietary codecs
+# (see chromium-gn.inc).
+PACKAGECONFIG[use-vaapi] = "use_vaapi=true,use_vaapi=false,libva"
+
 GN_ARGS += "\
         ${PACKAGECONFIG_CONFARGS} \
         use_ozone=true \
@@ -35,3 +43,11 @@ GN_ARGS += "use_x11=false"
 
 # The chromium binary must always be started with those arguments.
 CHROMIUM_EXTRA_ARGS_append = " --ozone-platform=wayland"
+
+addtask do_check_config before do_configure
+
+python do_check_config() {
+  if bb.utils.contains('PACKAGECONFIG', 'use-vaapi', True, False, d) and \
+    not bb.utils.contains('PACKAGECONFIG', 'proprietary-codecs', True, False, d):
+        bb.fatal("use-vaapi requires the proprietary-codecs option to be set")
+}
--- a/recipes-browser/chromium/README
+++ b/recipes-browser/chromium/README
@@ -75,6 +75,11 @@ PACKAGECONFIG knobs
   are desirable on embedded platforms. With this packageconfig, EGL and OpenGL
   ES 2.x are used instead.
 
+* use-vaapi: (off by default, chromium-ozone-wayland only)
+  Enables VA-API support with Ozone/Wayland backend that allows hardware accelerated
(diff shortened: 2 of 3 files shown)


recipes-browser/chromium/chromium-ozone-wayland_87.0.4280.141.bb
```diff
@@ -26,6 +26,17 @@ GN_ARGS += "\
         use_gtk=false \
 "
 
+SRC_URI += "\
+        file://0001-ozone-add-va-api-support-to-wayland.patch \
+"
+
+# Enables VA-API for Ozone/Wayland. Remember to also use proprietary codecs
+# (see chromium-gn.inc).
+PACKAGECONFIG[use_vaapi] = "use_vaapi=true,use_vaapi=false,libva"
```
rakuco, line 35: Can you document this new option in `README`? Also, looking at `chromium-gn.inc` it looks like we use the "foo-bar" naming format rather than "foo_bar", so it's better to call this `use-vaapi`.

recipes-browser/chromium/chromium-ozone-wayland_87.0.4280.141.bb
```diff
@@ -26,6 +26,17 @@ GN_ARGS += "\
         use_gtk=false \
 "
 
+SRC_URI += "\
+        file://0001-ozone-add-va-api-support-to-wayland.patch \
+"
+
+# Enables VA-API for Ozone/Wayland. Remember to also use proprietary codecs
```
rakuco, line 33: What happens when `use_vaapi` is enabled but `proprietary-codecs` is not? It's been a long time since I had to deal with `PACKAGECONFIG`, so I don't remember if there's a way to make one option depend on another.

recipes-browser/chromium/chromium-ozone-wayland_87.0.4280.141.bb
msisov: done

recipes-browser/chromium/chromium-ozone-wayland_87.0.4280.141.bb
msisov: Then hw video decoding won't work as proprietary-codecs enable h.264 code support. I could probably make packageconfig append proprietary-codecs if use-vaapi is set, but it will require chromium-ozone-wayland*.bb to be resolved before the main file from our recipe (chromium-gn.inc) is resolved.

Instead, I added a task that checks that proprietary codecs is set. if not, a fatal error is raised.

recipes-browser/chromium/chromium-ozone-wayland_87.0.4280.141.bb
```diff
@@ -35,3 +43,11 @@ GN_ARGS += "use_x11=false"
 
 # The chromium binary must always be started with those arguments.
 CHROMIUM_EXTRA_ARGS_append = " --ozone-platform=wayland"
+
+addtask do_check_config before do_configure
+
+python do_check_config() {
+  if bb.utils.contains('PACKAGECONFIG', 'use-vaapi', True, False, d) and \
+    not bb.utils.contains('PACKAGECONFIG', 'proprietary-codecs', True, False, d):
+        bb.fatal("ERROR: use-vaapi requires proprietary-codecs set!")
```
rakuco, line 52: I'd remove the exclamation mark and say "use-vaapi requires the proprietary-codecs option to be set" instead.

rakuco (approved):
lgtm with a nit

recipes-browser/chromium/chromium-ozone-wayland_87.0.4280.141.bb
msisov: done

rakuco (approved):
lgtm, kiitos!

PR #438, opened 2021-01-22

### LINKED ISSUE #431: Missing libatomic.so.1.2.0 (by native clang) in Centos8

hongxu-jia:
Since chromium uses clang as toolchain and clang delegates atomics to
runtime library instead of builtins, Link with latomic in TARGET_LDFLAGS
rather than in BUILD.gn, to avoid build failure due to no libatomic on
host (such as CentOS 8)

Fixes #431

### DIFF
--- a/recipes-browser/chromium/files/0001-remove-link-to-libatomic-from-BUILD.gn.patch
+++ b/recipes-browser/chromium/files/0001-remove-link-to-libatomic-from-BUILD.gn.patch
@@ -1,0 +1,61 @@
+Upstream-Status: Inappropriate [oe specific]
+
+Signed-off-by: Hongxu Jia <hongxu.jia@windriver.com>
+---
+
+From 01c899455bff7ab0322542c832d4f7c3359e9139 Mon Sep 17 00:00:00 2001
+From: Hongxu Jia <hongxu.jia@windriver.com>
+Date: Thu, 21 Jan 2021 17:07:28 +0800
+Subject: [PATCH] remove link to libatomic from BUILD.gn
+
+In BUILD.gn, the compile option does not distinguish target and native, such as
+-latomic works for both of target and native, if no libatomic on host (such as
+CentOS 8), there will be a link failure on native tool, even though the native
+tool does not require -latomic.
+
+Remove -latomic from BUILD.gn, and we will set it for target from recipe
+---
+ base/BUILD.gn               | 9 ---------
+ build/config/linux/BUILD.gn | 6 ------
+ 2 files changed, 15 deletions(-)
+
+diff --git a/base/BUILD.gn b/base/BUILD.gn
+index 1d97f72..ba755fb 100644
+--- a/base/BUILD.gn
++++ b/base/BUILD.gn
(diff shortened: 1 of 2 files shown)


recipes-browser/chromium/files/0001-remove-link-to-libatomic-from-BUILD.gn.patch
```diff
@@ -0,0 +1,61 @@
...
+@@ -27,12 +27,6 @@ config("runtime_library") {
+   if (is_chromeos) {
+     defines = [ "OS_CHROMEOS" ]
+   }
+-
+-  if ((!(is_chromeos || chromeos_is_browser_only) ||
+-       default_toolchain != "//build/toolchain/cros:target") &&
+-      (!use_custom_libcxx || current_cpu == "mipsel")) {
+-    libs = [ "atomic" ]
+-  }
```
rakuco, line 55: AFAICS this only applies to the target toolchain, right? Then you shouldn't need to remove this block (and change `chromium-gn.inc`).

recipes-browser/chromium/files/0001-remove-link-to-libatomic-from-BUILD.gn.patch
```diff
@@ -0,0 +1,61 @@
...
+   ]
+ 
+-  # Needed for <atomic> if using newer C++ library than sysroot, except if
+-  # building inside the cros_sdk environment - use host_toolchain as a
+-  # more robust check for this.
+-  if (!use_sysroot &&
+-      (is_android || ((is_linux || is_chromeos) && !is_chromecast)) &&
+-      host_toolchain != "//build/toolchain/cros:host") {
+-    libs += [ "atomic" ]
+-  }
```
rakuco, line 37: Rather than removing this block altogether, I guess you could just expand the check or replace the current `host_toolchain` one with our own `//build/toolchain/yocto:yocto_native`.

rakuco (commented):
I've never used CentOS myself, but it looks like there are official libatomic RPMs available for it. Isn't just installing them on the host an option?

r1mikey:
> I've never used CentOS myself, but it looks like there are official libatomic RPMs available for it. Isn't just installing them on the host an option?

This is correct, and is what we do internally to support the native part of the build.

recipes-browser/chromium/files/0001-remove-link-to-libatomic-from-BUILD.gn.patch
hongxu-jia: @rakuco , I am afraid it works for both of target and native, it took me lots of time to find out these two places to remove

recipes-browser/chromium/files/0001-remove-link-to-libatomic-from-BUILD.gn.patch
hongxu-jia: @rakuco Got it, I will have a try

PR #439, opened 2021-01-22
chromium-x11: install ANGLE so files (libEGL.so, libGLESv2.so)

hongxu-jia:
Refer what fedora [1] did, install libEGL.so, libGLESv2.so which
avoid failed to load libGLESv2.so.2 on bcm-2xxx-rpi4 when launch
chromium with cmd line

...
sh-5.0# chromium --no-sandbox
...

[1] https://src.fedoraproject.org/rpms/chromium/c/bbb6fbf8e77f620502fc655619b867f0aeb18f03.patch

### DIFF
--- a/recipes-browser/chromium/chromium-gn.inc
+++ b/recipes-browser/chromium/chromium-gn.inc
@@ -426,6 +426,9 @@ do_install() {
 	# modifying the dummy "CHROME_EXTRA_ARGS" line
 	sed -i "s/^CHROME_EXTRA_ARGS=\"\"/CHROME_EXTRA_ARGS=\"${CHROMIUM_EXTRA_ARGS}\"/" ${D}${libdir}/chromium/chromium-wrapper
 
+	# This is ANGLE, not to be confused with the similarly named files under swiftshader/
+	cp -a libEGL.so* libGLESv2.so* ${D}${libdir}/chromium/
+
 	if [ -n "${@bb.utils.contains('PACKAGECONFIG', 'component-build', 'component-build', '', d)}" ]; then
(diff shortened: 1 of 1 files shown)


rakuco (commented):
Thanks for the patch, it makes sense to me. I'd like @msisov to take a look since it affects chromium-ozone-wayland as well.

My only suggestion is to tweak the PR and commit message: instead of implying that we're just following Fedora without much explanation, it is better to point to the upstream change (https://chromium-review.googlesource.com/c/chromium/src/+/2044054) that did this and the bug it fixed (https://crbug.com/1049449).

msisov:
we don't use angle with ozone yet. what will happen if those shared libraries are missing?

rakuco:
> what will happen if those shared libraries are missing?

Do you mean if they are missing when we try to install them or at runtime? I _think_ they are always built but I haven't checked, at runtime I'd expect something like the upstream bug (https://crbug.com/1049449) maybe.

msisov:
My only concern is the installation step. I'm not sure if we have dependency on Angle at all in ozone only builds.

rakuco:
@hongxu-jia can you either make sure these files are also built by chromium-ozone-wayland or conditionally install them only if they are present?

hongxu-jia:
> @hongxu-jia can you either make sure these files are also built by chromium-ozone-wayland or conditionally install them only if they are present?

@rakuco No problem, I prefer to conditionally install them only if they are present, and I will resubmit the pull request

rakuco:
@hongxu-jia Thanks. Note that there's no reason to close this PR and send a new one, you can just update your commit and force-push it.

PR #441, opened 2021-01-22

### LINKED ISSUE #431: Missing libatomic.so.1.2.0 (by native clang) in Centos8

hongxu-jia:
When host (such as CentOS 8) did not install libatomic, there was a
link failure on native. In fact, only target requires to link atomic,
the native does not. So link atomic for target only by
(current_toolchain != host_toolchain)

Fixes #431

### DIFF
--- a/recipes-browser/chromium/chromium-gn.inc
+++ b/recipes-browser/chromium/chromium-gn.inc
@@ -22,6 +22,7 @@ SRC_URI += " \
         file://0001-drop-dep-on-ui-ozone-for-building-without-ozone.patch \
         file://0001-Fix-local-build.patch \
         file://0001-IWYU-ui-CursorFactory-is-required-without-Ozone.patch \
+        file://0001-link-atomic-for-target-only.patch \
 "
 
 SRC_URI_append_libc-musl = "\
--- a/recipes-browser/chromium/files/0001-link-atomic-for-target-only.patch
+++ b/recipes-browser/chromium/files/0001-link-atomic-for-target-only.patch
@@ -1,0 +1,52 @@
+Upstream-Status: Inappropriate [oe specific]
+
+Signed-off-by: Hongxu Jia <hongxu.jia@windriver.com>
+---
+From 83e2c76191d339719094e76f8a235c5192aa49ad Mon Sep 17 00:00:00 2001
+From: Hongxu Jia <hongxu.jia@windriver.com>
+Date: Fri, 22 Jan 2021 21:53:18 +0800
+Subject: [PATCH] link atomic for target only
+
+When host (such as CentOS 8) did not install libatomic, there was a
+link failure on native. In fact, only target requires to link atomic,
+the native does not. So link atomic for target only
+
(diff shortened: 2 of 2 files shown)


rakuco:
I maintain the question I asked in #438:
> I've never used CentOS myself, but it looks like there are official libatomic RPMs available for it. Isn't just installing them on the host an option?

yifan19:
@rakuco  Windriver wants to minimize the host requirement.

So we basically don't want to add any packages that are not listed in the yocto reference manual
https://www.yoctoproject.org/docs/latest/ref-manual/ref-manual.html#centos-8-packages

rakuco:
I see. Isn't that a matter of internal process, in which case the patch should be kept downstream?

I'm not categorically rejecting this patch or saying it shouldn't maybe be fixed upstream, but I'm still trying to understand the needs here -- if `libatomic` was in that specific document would it be OK to install it on the host? What if it was in this recipe's README?

One other thing that came to mind: does Yocto provide some libatomic-native recipe we could depend on? We already use clang-native rather than the system compiler from https://www.yoctoproject.org/docs/latest/ref-manual/ref-manual.html#centos-8-packages anyway, so if that recipe exists we could also just add another -native dependency.

hongxu-jia:
> I maintain the question I asked in #438:
> 
> > I've never used CentOS myself, but it looks like there are official libatomic RPMs available for it. Isn't just installing them on the host an option?

I think it is another question:

What are the Build Host Packages required by meta-browser,
the Yocto project has documented in its quick start [1] and
no libatomic required on centos

[1] https://www.yoctoproject.org/docs/2.4.2/yocto-project-qs/yocto-project-qs.html#packages

If no documentation, the user have to install it after a failure
which is poor user experience; further more, install a package on
host usually requires root permission which is also an inconvenience

For this defect, I think we should link atomic at where it is used, not everywhere

hongxu-jia:
> @rakuco Windriver wants to minimize the host requirement.
> 
> So we basically don't want to add any packages that are not listed in the yocto reference manual
> https://www.yoctoproject.org/docs/latest/ref-manual/ref-manual.html#centos-8-packages
> 
> to my surprise, centos-8 doesn't have libatomic coming with libc or gcc

The clang delegates atomics to runtime library so link,
and gcc has atomics builtins so no link

In CentOS 8, it seems not friendly to clang

yifan19:
> One other thing that came to mind: does Yocto provide some libatomic-native recipe we could depend on? We already use clang-native rather than the system compiler from https://www.yoctoproject.org/docs/latest/ref-manual/ref-manual.html#centos-8-packages anyway, so if that recipe exists we could also just add another -native dependency.

hongxu probably knows more than me...

Here is what I know:

I remember Randy Macleod (Windriver) asking this in the last Yocto bug meeting to Richard Purdie about "building libatomic" and Richard said no.

I think in order to build libatomic-native, we need to build most of gcc-toolchain. There isn't a recipe for libatomic on the oe-core index.

PR #442, opened 2021-01-22
chromium-x11: install ANGLE so files (libEGL.so, libGLESv2.so)

hongxu-jia:
Refer what fedora [1] did, install libEGL.so, libGLESv2.so which
avoid failed to load libGLESv2.so.2 on bcm-2xxx-rpi4 when launch
chromium with cmd line

...
sh-5.0# chromium --no-sandbox
...

[1] https://src.fedoraproject.org/rpms/chromium/c/bbb6fbf8e77f620502fc655619b867f0aeb18f03.patch

### DIFF
--- a/recipes-browser/chromium/chromium-gn.inc
+++ b/recipes-browser/chromium/chromium-gn.inc
@@ -426,6 +426,14 @@ do_install() {
 	# modifying the dummy "CHROME_EXTRA_ARGS" line
 	sed -i "s/^CHROME_EXTRA_ARGS=\"\"/CHROME_EXTRA_ARGS=\"${CHROMIUM_EXTRA_ARGS}\"/" ${D}${libdir}/chromium/chromium-wrapper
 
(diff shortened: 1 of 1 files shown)


rakuco:
From #439:
> My only suggestion is to tweak the PR and commit message: instead of implying that we're just following Fedora without much explanation, it is better to point to the upstream change (https://chromium-review.googlesource.com/c/chromium/src/+/2044054) that did this and the bug it fixed (https://crbug.com/1049449).

You also need to s/chromium-x11/chromium/, as your change affects both recipes.

hongxu-jia:
> From #439:
> 
> > My only suggestion is to tweak the PR and commit message: instead of implying that we're just following Fedora without much explanation, it is better to point to the upstream change (https://chromium-review.googlesource.com/c/chromium/src/+/2044054) that did this and the bug it fixed (https://crbug.com/1049449).
> 
> You also need to s/chromium-x11/chromium/, as your change affects both recipes.

Got it

PR #444, opened 2021-01-22
chromium: install ANGLE so files (libEGL.so, libGLESv2.so)

hongxu-jia:
Since upstream [Ship ANGLE libEGL.so and libGLESv2.so on Linux stable][1]
to fix [WebGL seems to be disabled in Pixi JS on Chrome 80][2]. Do the
same thing in recipe, otherwise there was a failure when launch
chromium with command line on target raspberry pi 4:
...
sh-5.0# chromium --no-sandbox
...

[1] https://chromium-review.googlesource.com/c/chromium/src/+/2044054
[2] https://crbug.com/1049449)


rakuco (approved):
lgtm. In the future, please make your changes in the same pull request rather than closing and submitting new ones, this is very confusing (besides being unnecessary) -- you just need to amend your commits and force-push.

PR #448, opened 2021-01-27
chromium: Backport patch to stop including <sys/random.h>

### LINKED ISSUE #445: chromium: Native builds broken on CentOS 7
Just an informational note.

The native component of the Chromium build is broken on CentOS 7 (and possibly other older kernels) due to a missing `sys/random.h`.  I first noticed this with the 87.x update.  We work around this locally with the patch below, which simply includes `linux/random.h` instead.

I suspect that the best approach might be to simply document in the README that CentOS older than version 8 is not supported as a build host (given that one would need at least the `devtoolset-7` package set on the older OS anyway).

```
--- a/base/rand_util_posix.cc	2021-01-22 07:18:29.929351787 -0800
+++ b/base/rand_util_posix.cc	2021-01-22 07:18:36.139314167 -0800
@@ -24,7 +24,7 @@
 #if !defined(OS_IOS) && !defined(OS_NACL)
 // TODO(crbug.com/995996): Waiting for this header to appear in the iOS SDK.
 // (See below.)
-#include <sys/random.h>
+#include <linux/random.h>
 #endif

 namespace {
```

rakuco:
This header might not exist on systems with very old glibc packages (e.g.
CentOS 7) when building some host binaries that we need to run as part of
the build. It turns out the header is not necessary at all, as Chromium uses
LSS (Linux Syscall Support) to call the syscall and falls back to reading
from /dev/urandom when that fails.

Fixes #445.

### DIFF
--- a/recipes-browser/chromium/chromium-gn.inc
+++ b/recipes-browser/chromium/chromium-gn.inc
@@ -22,6 +22,7 @@ SRC_URI += " \
         file://0001-drop-dep-on-ui-ozone-for-building-without-ozone.patch \
         file://0001-Fix-local-build.patch \
         file://0001-IWYU-ui-CursorFactory-is-required-without-Ozone.patch \
+        file://0001-RandBytes-Stop-including-sys-random.h-on-Linux.patch \
 "
 
 SRC_URI_append_libc-musl = "\
--- a/recipes-browser/chromium/files/0001-RandBytes-Stop-including-sys-random.h-on-Linux.patch
+++ b/recipes-browser/chromium/files/0001-RandBytes-Stop-including-sys-random.h-on-Linux.patch
@@ -1,0 +1,49 @@
+Upstream-Status: Backport
+
+Signed-off-by: Raphael Kubo da Costa <raphael.kubo.da.costa@intel.com>
+---
+From 80fb001240ed0c30f06eabc75a7899bf1067f515 Mon Sep 17 00:00:00 2001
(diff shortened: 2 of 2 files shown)


recipes-browser/chromium/chromium-gn.inc
```diff
@@ -22,6 +22,7 @@ SRC_URI += " \
         file://0001-drop-dep-on-ui-ozone-for-building-without-ozone.patch \
         file://0001-Fix-local-build.patch \
         file://0001-IWYU-ui-CursorFactory-is-required-without-Ozone.patch \
+        file://0001-RandBytes-Stop-including-sys-random.h-on-Linux.patch \
```
msisov, line 25: check spaces.

msisov (changes requested):

recipes-browser/chromium/chromium-gn.inc
rakuco: Looks fine to me, isn't it using the same indentation as the other lines?

msisov: nope. for me, it shows like it has 1 extra space.

rakuco: Hmm, I still don't see it. The "f" from "file://" looks perfectly aligned under the other lines, and I don't see any trailing white space either.

msisov: Ok, let's try to merge it and see if it looks ok after that.

msisov (approved):

PR #449, opened 2021-01-27
chromium: avoid link latomic failure on CentOS 8 host

### LINKED ISSUE #431: Missing libatomic.so.1.2.0 (by native clang) in Centos8

yifan19:
replaces #443 
----
When host (such as CentOS 8) did not install libatomic, there was a
link failure on native. In fact, only target requires to link atomic,
the native does not. So link atomic for target only by
(current_toolchain != host_toolchain)

Fixes #431

### DIFF
--- a/recipes-browser/chromium/chromium-gn.inc
+++ b/recipes-browser/chromium/chromium-gn.inc
@@ -24,6 +24,7 @@ SRC_URI += " \
         file://0001-IWYU-add-missing-include-for-std-vector-and-std-uniq.patch \
         file://0001-ozone-fix-include.patch \
         file://0001-Fix-ill-formed-C-code.patch \
+        file://0001-link-atomic-for-target-only.patch \
 "
 
 SRC_URI_append_libc-musl = "\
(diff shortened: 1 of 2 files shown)


kraj (approved):
I think its not host vs target, but rather atomics support is available or not. Since our build hosts are x86/aarch64 for now, this patch should suffice to fix the problem. Ideally a proper check would be better to identify where atomics are implemented ( intrinsics or libatomic etc. )

yifan19:
i was testing it after applying this patch and it didn't solve the problem, host clang still included -latomic

yifan19:
test passed for intel86_64, after looking at logs I believe 
the name the toolchains //build/toolchain/yocto:yocto_target  //build/toolchain/yocto:yocto_native

rakuco:
I'd like more time to do some testings here too. Upstream seems to be too eager to link against libatomic even when not necessary, but I'd like to double-check what we're currently doing with libc++ and compiler-rt, otherwise we risk just masking the issue if/when we end up with native binaries that do need atomics support.

yifan19:
> I'd like more time to do some testings here too. Upstream seems to be too eager to link against libatomic even when not necessary, but I'd like to double-check what we're currently doing with libc++ and compiler-rt, otherwise we risk just masking the issue if/when we end up with native binaries that do need atomics support.

@rakuco can this patch be merged? have you taken a look at libc++ and compiler-rt?

rakuco (approved):
let's merge it, I don't think I'll have time to investigate this any time soon and it's not wrong per se.

PR #457, opened 2021-02-05
chromium: backport patch that addresses cve-2021-21148 (0day)

msisov:
A stable channel has been updated to 88.0.4324.150, but it hasn't
been rolled out yet. Thus, backport the patch and keep it until
the meta-browser chromium recipe is updated to 88.0.4324.150.


rakuco (approved):
lgtm, thanks for tracking this

PR #458, opened 2021-02-08
chromium: specify the current supported architectures

yifan19:
These following architectures have been successfully
tested.

* arm64
* arm32
* x86-64


rakuco:
We already set `COMPATIBLE_MACHINE` earlier in the same file, could you clarify how setting this new variable improves the situation over changing the former instead?

yifan19:
Sorry I totally overlooked that.
After quick research, seems the flag do essentially the same thing.

PR #482, opened 2021-03-20
chromium: Use libc++/compiler-rt for C/C++ runtime

kraj:
Lets switch to using libc++ instead of libstdc++ as that's the default chromium uses. Infact it uses a vendored version
but we disable it as of now. https://github.com/OSSystems/meta-browser/blob/master/meta-chromium/recipes-browser/chromium/chromium-gn.inc#L160
that's perhaps another step to start using that but it can go out of sync with clang provided versions. but we should try

### DIFF
--- a/meta-chromium/recipes-browser/chromium/chromium-gn.inc
+++ b/meta-chromium/recipes-browser/chromium/chromium-gn.inc
@@ -477,3 +477,5 @@ PACKAGE_DEBUG_SPLIT_STYLE = "debug-without-src"
 
 # There is no need to ship empty -dev packages.
 ALLOW_EMPTY_${PN}-dev = "0"
+
+RUNTIME = "llvm"


meta-chromium/recipes-browser/chromium/files/0014-sql-Disable-is_standard_layout-assert.patch
```diff
@@ -0,0 +1,32 @@
...
+From: Khem Raj <raj.khem@gmail.com>
+Date: Fri, 19 Mar 2021 15:36:35 -0700
+Subject: [PATCH] sql: Disable is_standard_layout assert
+
+This does not work with GCC 11 runtime
+fixes
+In file included from ../../sql/recover_module/cursor.cc:5:
+../../sql/recover_module/cursor.h:65:5: error: static_assert failed due to requirement 'std::is_standard_layout<sql::recover::VirtualCursor>::value' "needed for the reinterpret_cast below"
+    static_assert(std::is_standard_layout<VirtualCursor>::value,
+
```
otavio, line 11: Please add the upstream status information and mentioned the bug report here so it can be referred when updating the recipe.

otavio (changes requested):

meta-chromium/recipes-browser/chromium/files/0014-sql-Disable-is_standard_layout-assert.patch
otavio: Shouldn't the referred bug report to be included in the commit log?

kraj: done

otavio (approved):

rakuco:
By GCC runtime do you mean libstdc++ or is this something else?

kraj:
> By GCC runtime do you mean libstdc++ or is this something else?
yes, but C++ compiler headers in this case particularly, I think all linux distros will run into this problem once they start
using gcc 11 and try to compile chromium.

rakuco:
Weren't we using libc++ in the build? I'm still trying to understand if it's the standard C++ library or some header that will always come from GCC regardless of compiler/standard library choice.

kraj:
> Weren't we using libc++ in the build? I'm still trying to understand if it's the standard C++ library or some header that will always come from GCC regardless of compiler/standard library choice.

as I mentioned these are compiler headers, if you are more curious look at the preprocessed files I have attached to
https://bugs.chromium.org/p/chromium/issues/detail?id=1189788
you will understand more.

rakuco:
OK, AFAICS the preprocessed file confirmed that it's libstdc++, which means we're not using libc++ as I thought we already were. I think we have 2 options: switch to libc++ or wait for a response from upstream for this issue, as I think removing this `static_assert()` call will make the code build but crash at runtime (based on what the assert is checking).

kraj:
> OK, AFAICS the preprocessed file confirmed that it's libstdc++, which means we're not using libc++ as I thought we already were. I think we have 2 options: switch to libc++ or wait for a response from upstream for this issue, as I think removing this `static_assert()` call will make the code build but crash at runtime (based on what the assert is checking).

yeah I think using libc++ is the right thing, although that means there are two options, user internal vendored version of libc++ which maybe out of sync with clang used to compile it. Or use `RUNTIME = "llvm"` in local.conf but that means now you are switching system defaults to use libc++ instead of libstdc++, I think option 1 is good.

rakuco:
By option 1 do you mean switching to Chromium's bundled libc++ and compiler_rt copy? If we go for the other option, isn't it safe to just set `RUNTIME` to "llvm" in `chromium.inc` like we do for `TOOLCHAIN = "clang"`?

kraj:
> By option 1 do you mean switching to Chromium's bundled libc++ and compiler_rt copy? If we go for the other option, isn't it safe to just set `RUNTIME` to "llvm" in `chromium.inc` like we do for `TOOLCHAIN = "clang"`?

hmm thats good option to make it at recipe level

kraj:
@rakuco I have switched chromium to use llvm runtime, and it fixes the problem I was observing.

rakuco:
I'm testing some builds with this setting here too. It's a big change, so my plan is to file an issue, get @msisov and others to test it too and then make the switch in another PR.

kraj:
> I'm testing some builds with this setting here too. It's a big change, so my plan is to file an issue, get @msisov and others to test it too and then make the switch in another PR.

OK sounds good. I will keep this around until then and close it after referring the future PR

rakuco:
> so my plan is to file an issue

#488, for the record.

kraj:
Please test this change together with https://github.com/OSSystems/meta-browser/pull/489 or wait for that PR to land in master.  if you don't have a lot of RAM otherwise you might see linker crashes which are not related to this change but to layer restructuring.

yifan19:
> Please test this change together with #489 or wait for that PR to land in master. if you don't have a lot of RAM otherwise you might see linker crashes which are not related to this change but to layer restructuring.

as discussed recently #485, segfault can be seen even if you have around 126GB of RAM

kraj:
> > Please test this change together with #489 or wait for that PR to land in master. if you don't have a lot of RAM otherwise you might see linker crashes which are not related to this change but to layer restructuring.
> 
> as discussed recently #485, segfault can be seen even if you have around 126GB of RAM

hmm that means there is something going on with gold itself. I don't plan to debug it myself.

rakuco:
Replaced by #492.

PR #492, opened 2021-03-31
chromium: Switch to using libc++ and compiler_rt in target builds.

### LINKED ISSUE #488: chromium: Switch to building with libc++
This was originally discussed in #482, which covered a build issue with the upcoming libstdc++ from GCC 11.

I thought we had already been using libc++ ever since we started building with clang in #290, but apparently we only switched compilers back then. libstdc++ is not officially supported or tested upstream, so it is up to the community to send patches when things break, and the issue in #482 is another example. Switching to libc++ would get us closer to what upstream builds and tests and allow us to remove a few patches we currently carry.

Things to consider:
* Whether to use Chromium's own bundled libc++ and compiler_rt copy. Doing so would get us even closer to what upstream builds and ships, but using bundled libraries is something frowned upon by OE and Yocto and there may be a risk that things don't work when mixing them with the older clang from meta-clang ("older" compared to the upstream LLVM snapshot that Chromium ships and uses as a compiler).
* What to do with the native parts of the build. Setting `RUNTIME = "llvm"` in `chromium.inc` does not cause libc++ to be used for the native parts of the build, so we still risk running into libstdc++ build issues. Not sure if this is something that needs to be fixed in meta-clang.
* Testing. I've been doing builds with `RUNTIME = "llvm"` here, but help running the builds and doing more builds under other settings is also appreciated.

rakuco:
This follows what upstream has done in its official builds for years. We had
already been using clang in our builds, but with libstdc++. This has
occasionally caused a number of build issues over the years (e.g. #482), so
it makes sense to just follow upstream's toolchain choices more closely.

Due to https://github.com/kraj/meta-clang/issues/449 the native parts of the
build are still using libstdc++ though, so not all libstdc++-related patches
could be removed in this commit.

This change has been tested exhaustively by multiple people as reported in
the bug below. Big thanks to @kraj, @msisov, @MastaG and @r1mikey for their
testing!

Fixes #488.


kraj (changes requested):
I think 0013-mojo-Include-string.h-for-strncpy-declaration.patch is not related to libc++/libstdc++  switch, its a normal bug where source file uses strncpy() API but does not include other needed header file,  I think it should be upstreamed if its not already in. Removing other two is ok

rakuco:
It's been upstreamed. Like `0012-IWYU-add-ctime-for-std-time.patch` it's explicitly including a header that is used in the file, but just like it the build works fine with libc++ (which is why it had to be fixed upstream in the first place).

kraj (approved):

rakuco:
Thanks!

Issue #473, opened 2021-03-17
Chromium: missing font display when running without --no-sandbox on qemux86-64


yifan19:
built with latest poky + meta-clang

MACHINE: qemux86-64
bitbake core-image-sato

```
runqemu qemux86-64 kvm qemuparams="-m 4000 -display sdl -vga std" serialstdio
```

run `DISPLAY=:0.0 chromium 1.1.1.1` as a user and you see no font display, image works ok.
https://imgur.com/a/lITEEKE

runs with --no-sandbox as user fixes the problem (the error/warning msg stay the same)

error msg (doesn't impact browsing)
```
^Cqemux86-64:~$ DISPLAY=:0.0 chromium 1.1.1.1  
[1932:1932:0317/143707.902184:ERROR:browser_dm_token_storage_linux.cc(94)] Error: /etc/machine-id contains 0 characters (32 were expected).
[1963:1963:0317/143707.915188:ERROR:gl_implementation.cc(283)] Failed to load libGLESv2.so.2: libGLESv2.so.2: cannot open shared object file: No such file or directory
[1963:1963:0317/143707.927250:ERROR:viz_main_impl.cc(150)] Exiting GPU process due to errors during initialization
[1932:1961:0317/143707.953809:ERROR:bus.cc(393)] Failed to connect to the bus: Could not parse server address: Unknown address type (examples of valid types are "tcp" and on UNIX "unix")
[1932:1961:0317/143707.953841:ERROR:bus.cc(393)] Failed to connect to the bus: Could not parse server address: Unknown address type (examples of valid types are "tcp" and on UNIX "unix")
[1932:1961:0317/143707.997092:ERROR:bus.cc(393)] Failed to connect to the bus: Could not parse server address: Unknown address type (examples of valid types are "tcp" and on UNIX "unix")
[1932:1961:0317/143707.998560:ERROR:bus.cc(393)] Failed to connect to the bus: Could not parse server address: Unknown address type (examples of valid types are "tcp" and on UNIX "unix")
[1932:2016:0317/143708.166145:ERROR:object_proxy.cc(621)] Failed to call method: org.freedesktop.DBus.Properties.Get: object_path= /org/freedesktop/UPower: org.freedesktop.DBus.Error.ServiceUnknown: The name org.freedesktop.UPower was not provided by any .service files
[...]

PR #495, opened 2021-04-06
Fix font rendering with glibc-2.33

yifan19:
A change in the fstat syscall in glibc-2.33
causes font rendering issues when sandbox is enabled.

https://github.com/OSSystems/meta-browser/issues/473

### DIFF
--- a/meta-chromium/recipes-browser/chromium/chromium-gn.inc
+++ b/meta-chromium/recipes-browser/chromium/chromium-gn.inc
@@ -19,6 +19,7 @@ SRC_URI += " \
     file://0009-RandBytes-Stop-including-sys-random.h-on-Linux.patch \
     file://0010-avoid-link-latomic-failure-on-CentOS-8-host.patch \
     file://0011-Fix-use-of-DCHECK-with-std-unique_ptr.patch \
+    file://0012-Fix-font-rendering-with-glibc-2.33.patch \
 "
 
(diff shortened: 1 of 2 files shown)


otavio (changes requested):
The change itself could be ordered properly in `SRC_URI`; also, the commit log should use `chromium: ` prefix.

yifan19:
note: qt5-webengine might also face the same issue

otavio (approved):

kraj (approved):
yep LGTM

otavio:
@shr-project maybe this affects QtWebEngine as well?

shr-project:
qtwebengine-5.15.2 in meta-qt5 is based on 83 chromium and I have some patches for 5.15.3 with 87 chromium, but don't plan to merge them to meta-qt5/master due to license change in 5.15.3 LTS to be commercial only https://www.qt.io/blog/commercial-lts-qt-5.15.3-released

So backporting anything which exists in newer qtwebengine-chromium.git might be complicated due to license, but I don't know how TQC is going to deal with the license of chromium bits (and 87-based branch of qtwebengine-chromium.git as well as 5.15.3 upgrade of qtwebengine.git is publicly available while e.g. qtbase 5.15.3 is only available for their paying customers).

Anyway I don't plan to spend much time on meta-qt5, it's dead-end anyway (and old 83* chromium there bothers me more than broken fonts with glibc-2.33).

PR #498, opened 2021-04-14
chromium: Add opengl to REQUIRED_DISTRO_FEATURES

yifan19:
Chromium depends on virtual/libgl,
which oe-core provides by default in mesa.

opengl or vulkan is a REQUIRED_DISTRO_FEATURES for mesa.

In a minimal build of oe-core + meta-oe + meta-python2,
bitbake would complain that nothing provides
virtual/libgl, virtual/libgles2.

Also makes meta-chromium pass yocto-check-layer.

### DIFF
--- a/meta-chromium/recipes-browser/chromium/chromium-gn.inc
+++ b/meta-chromium/recipes-browser/chromium/chromium-gn.inc
@@ -45,6 +45,8 @@ SRC_URI_append_libc-musl = "\
     file://musl/0020-Fix-tab-crashes-on-musl.patch \
 "
 
+ANY_OF_DISTRO_FEATURES = "opengl vulkan"
+
 # Append instead of assigning; the gtk-icon-cache class inherited above also
 # adds packages to DEPENDS.
 DEPENDS += " \


meta-chromium/recipes-browser/chromium/chromium-gn.inc
```diff
@@ -45,6 +45,8 @@ SRC_URI_append_libc-musl = "\
     file://musl/0020-Fix-tab-crashes-on-musl.patch \
 "
 
+REQUIRED_DISTRO_FEATURES = "opengl"
```
otavio, line 48: This should be `opengl`OR `vulkan`, right?

otavio (changes requested):

meta-chromium/recipes-browser/chromium/chromium-gn.inc
yifan19: sure that was the other alternative, following what mesa has.
I am not familiar with `vulkan`
will send v2

yifan19:
`vulkan` doesn't provide egl in recipe mesa though

ERROR: Nothing PROVIDES 'virtual/egl' (but /yow-lpggp32/yyu1/yocto/openembedded-core/meta-browser/meta-chromium/recipes-browser/chromium/chromium-x11_89.0.4389.114.bb DEPENDS on or otherwise requires it). Close matches:
  virtual/kernel
  virtual/mesa
  virtual/perf
ERROR: Required build target 'chromium-x11' has no buildable providers.
Missing or unbuildable dependency chain was: ['chromium-x11', 'virtual/egl']

yifan19:
it is buildable with vulkan if you remove PACKAGECONFIG `use-egl` and use `mesa-gl`

yifan19:
@otavio  can this be merged, i used ANY_OF_DISTRO instead

yifan19:
fixed commit log to  ANY_DISTRO_FEATURES

otavio (approved):

PR #501, opened 2021-04-21
WIP/RFC: chromium: move dbus to PACKAGECONFIG

yifan19:
dbus is a (r)dpendency of chromium as it tries to
communicate with different dbus services on launch.

When running it with a popular DE such as xfce
no issue occurs.

However, when trying to run a minimal embedded build,
(only Xserver, no DE & kiosk-mode as a service), chromium complains
about missing `upower` and missing environment variables such as
`XDG_RUNTIME_DIR`

This patch attempts to make the dependency to dbus optional
by reapplying a patch on chromium-85
from https://github.com/smaeul/portage-overlay/
    blob/master/www-client/chromium/files/chromium-85-optional-dbus.patch

### DIFF
--- a/meta-chromium/recipes-browser/chromium/chromium-gn.inc
+++ b/meta-chromium/recipes-browser/chromium/chromium-gn.inc
@@ -20,6 +20,7 @@ SRC_URI += " \
     file://0010-avoid-link-latomic-failure-on-CentOS-8-host.patch \
     file://0011-Fix-use-of-DCHECK-with-std-unique_ptr.patch \
     file://0012-Fix-font-rendering-with-glibc-2.33.patch \
+    file://0013-dbus-Make-dbus-optional.patch \
 "
 
 SRC_URI_append_libc-musl = "\
@@ -55,7 +56,6 @@ DEPENDS += " \
     atk \
     bison-native \
     cairo \
-    dbus \
     expat \
     flac \
     freetype \
@@ -98,7 +98,7 @@ BUILD_CC_toolchain-clang = "clang"
 BUILD_CXX_toolchain-clang = "clang++"
 BUILD_LD_toolchain-clang = "clang"
 
-PACKAGECONFIG ??= "use-egl"
+PACKAGECONFIG ??= "use-egl dbus"
 
 # this makes sure the dependencies for the EGL mode are present; otherwise, the configure scripts
 # automatically and silently fall back to GLX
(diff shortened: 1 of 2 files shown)


msisov:
IMO, this kind of changes must go to the upstream first. And the patch can be backported then. Otherwise, the recipe will keep a lot of patches that we will have to rebase and maintain. Of course, the patch can be refused to be accepted if the upstream is not interested, but I think this patch shouldn’t have any problems.

thanks!

yifan19:
> IMO, this kind of changes must go to the upstream first. And the patch can be backported then. Otherwise, the recipe will keep a lot of patches that we will have to rebase and maintain. Of course, the patch can be refused to be accepted if the upstream is not interested, but I think this patch shouldn’t have any problems.
> 
> thanks!

ok sending upstream too. thanks!

rakuco:
Also note that this is just the bare minimum you need to do to get the code to build, but I think there's a lot of code that ends up depending on dbus at runtime. If you grep for `dbus::` in the code base and take out all the ChromeOS-specific bits and pieces you'll still end up with code that expects a working D-Bus installation. As such, even if upstream takes some build fix patches, these code paths won't be tested very and you might have a somewhat broken Chromium binary.

It might be more useful to try to have D-Bus installed but address each issue individually. If you don't want to ship UPower, for example, try making it optional. Variables like `XDG_RUNTIME_DIR` aren't D-Bus specific and should probably be properly set anyway.

yifan19:
> Also note that this is just the bare minimum you need to do to get the code to build, but I think there's a lot of code that ends up depending on dbus at runtime. If you grep for `dbus::` in the code base and take out all the ChromeOS-specific bits and pieces you'll still end up with code that expects a working D-Bus installation. As such, even if upstream takes some build fix patches, these code paths won't be tested very and you might have a somewhat broken Chromium binary.

Yop definitely need to do more testing. I did a quick run in qemu and it seems to be functional.
if you do a grep for USE_DBUS, it was kind of supported even before this patch, but upstream doesn't seem to test this build option.

If it is currently broken, having it merged still means that we are the only few people maintaining it.

> 
> It might be more useful to try to have D-Bus installed but address each issue individually. If you don't want to ship UPower, for example, try making it optional. Variables like `XDG_RUNTIME_DIR` aren't D-Bus specific and should probably be properly set anyway.

agreed

yifan19:
reopen when there is a need for removing dbus, as suggested, upstream won't test this build path...

PR #502, opened 2021-04-21
chromium: Add UPower to RDPENDS

yifan19:
Chromium uses D-Bus to communicate with UPower.
It uses it to query battery among many things.
see:
    services/device/battery/battery_status_manager_linux.cc

Fixes error/warning messages related to UPower:
```
Failed to call method: org.freedesktop.UPower.GetDisplayDevice
The name org.freedesktop.UPower was not provided by any .service files
```

### DIFF
--- a/meta-chromium/README.md
+++ b/meta-chromium/README.md
@@ -117,6 +117,10 @@ PACKAGECONFIG knobs
   also be enabled. Please note that not all the possible hardware configs are
   tested and supported.
 
+* upower: (on by default)
+  Chromium expects the presence of `org.freedesktop.UPower` via D-Bus to
+  query battery status. If disabled, there will be warning messages seen on
+  stderr and Battery Status Web API will not work.
 
 Google API keys
 ---------------
--- a/meta-chromium/recipes-browser/chromium/chromium-gn.inc
+++ b/meta-chromium/recipes-browser/chromium/chromium-gn.inc
@@ -106,7 +106,7 @@ BUILD_CC:toolchain-clang = "clang"
 BUILD_CXX:toolchain-clang = "clang++"
 BUILD_LD:toolchain-clang = "clang"
 
-PACKAGECONFIG ??= "use-egl"
+PACKAGECONFIG ??= "upower use-egl"
 
 # this makes sure the dependencies for the EGL mode are present; otherwise, the configure scripts
 # automatically and silently fall back to GLX
@@ -123,6 +123,9 @@ PACKAGECONFIG[proprietary-codecs] = ' \
         ffmpeg_branding="Chrome" proprietary_codecs=true, \
         ffmpeg_branding="Chromium" proprietary_codecs=false \
 '
+# Enable UPower by default.
+# Chromium expects to be able to query battery status through D-Bus.
+PACKAGECONFIG[upower] = ",,,upower"
 
 # Disable VA-API by default. It is compile time enabled since M88, but it's not
 # desired to make all the users of the Chromium meta-browser recipe depend on


yifan19:
Note: I looked at Gentoo:
upower is not mentioned.
but dbus is mentioned as both DEPENDS and RDEPENDS
https://gitweb.gentoo.org/repo/gentoo.git/tree/www-client/chromium/chromium-90.0.4430.72.ebuild#n71

should this be a rrecommends then?

meta-chromium/recipes-browser/chromium/chromium-gn.inc
```diff
@@ -82,7 +82,10 @@ DEPENDS += " \
 "
 
 # The wrapper script we use from upstream requires bash.
-RDEPENDS_${PN} = "bash"
+RDEPENDS_${PN} = " \
+    bash \
+    upower \
```
otavio, line 87: I'd add it using a `PACKAGECONFIG` so people can choose to add it or not.

yifan19: ok

threexc:
Any updates on the review?

meta-chromium/recipes-browser/chromium/chromium-gn.inc
rakuco: @otavio should it really be _that_ optional? If UPower is not present, there's at least one Web API that will not work as expected. Is UPower not commonly shipped in embedded systems/is it too big?

meta-chromium/recipes-browser/chromium/chromium-gn.inc
```diff
@@ -116,6 +116,7 @@ PACKAGECONFIG[proprietary-codecs] = ' \
         ffmpeg_branding="Chrome" proprietary_codecs=true, \
         ffmpeg_branding="Chromium" proprietary_codecs=false \
 '
+PACKAGECONFIG[upower] = ",,,upower"
```
rakuco, line 119: If we do go with this approach, I think this should be on by default, and it needs to be documented in `README.md`.

meta-chromium/recipes-browser/chromium/chromium-gn.inc
otavio: If the application runs without it, it is optional. As suggested below, I think it should be enabled by default.

meta-chromium/recipes-browser/chromium/chromium-gn.inc
otavio: Yep. We should enable it but default.

otavio: You mentioned (on the other comment) that there are API which will fail. It might be good to mention it here as a comment in the recipe too.

otavio:
@yifan19 can you address @rakuco comments?

meta-chromium/README.md
```diff
@@ -117,6 +117,10 @@ PACKAGECONFIG knobs
   also be enabled. Please note that not all the possible hardware configs are
   tested and supported.
 
+* upower: (on by default)
+  Chromium expects the presence of `org.freedesktop.UPower` via the dbus to
```
rakuco, line 121: the dbus -> D-Bus

meta-chromium/README.md
```diff
@@ -117,6 +117,10 @@ PACKAGECONFIG knobs
...
+  query battery status. If disabled, there will be warning messages seen on
+  stderr.
```
rakuco, line 123: I'd also say that the Battery Status Web API will not work.

meta-chromium/recipes-browser/chromium/chromium-gn.inc
```diff
@@ -106,7 +106,7 @@ BUILD_CC:toolchain-clang = "clang"
 BUILD_CXX:toolchain-clang = "clang++"
 BUILD_LD:toolchain-clang = "clang"
 
-PACKAGECONFIG ??= "use-egl"
+PACKAGECONFIG ??= "use-egl upower"
```
rakuco, line 109: Can you keep this list sorted?

meta-chromium/recipes-browser/chromium/chromium-gn.inc
```diff
@@ -123,6 +123,9 @@ PACKAGECONFIG[proprietary-codecs] = ' \
         ffmpeg_branding="Chrome" proprietary_codecs=true, \
         ffmpeg_branding="Chromium" proprietary_codecs=false \
 '
+# Enable upower by default.
```
rakuco, line 126: upower -> UPower

meta-chromium/recipes-browser/chromium/chromium-gn.inc
```diff
@@ -123,6 +123,9 @@ PACKAGECONFIG[proprietary-codecs] = ' \
...
+# Chromium expects to be able to query battery status through dbus.
```
rakuco, line 127: dbus -> D-Bus

rakuco (approved):
Thanks, this looks good. I've only made some suggestions to the comments, and after that we're good to go!

rakuco (approved):
lgtm, thanks!

PR #504, opened 2021-04-28
README: add meta-browser typical use case

kenhys:
Adding a typical use case may help the user to understand it.

### DIFF
--- a/meta-chromium/README.md
+++ b/meta-chromium/README.md
@@ -37,7 +37,7 @@ $ git clone git://github.com/OSSystems/meta-browser.git
 $ source poky/oe-init-build-env
 $ bitbake-layers add-layer ../meta-openembedded/meta-oe ../meta-clang ../meta-python2 ../meta-browser/meta-chromium
 ```
-And, append target as I in conf/local.conf.
+And, append target in conf/local.conf.
 
 ```sh
 IMAGE_INSTALL_append = " chromium-ozone-wayland "
--- a/meta-chromium/README.md
+++ b/meta-chromium/README.md
@@ -23,6 +23,28 @@ This layer depends on:
   - branch: master
   - revision: HEAD
 
+Usage
+-----
+
+The following steps are typical use case which builds chromium-ozone-wayland for example.
+
+```sh
+$ git clone git://git.yoctoproject.org/poky
+$ git clone git://git.openembedded.org/meta-openembedded
+$ git clone git://github.com/kraj/meta-clang
+$ git clone git://git.openembedded.org/meta-python2
+$ git clone git://github.com/OSSystems/meta-browser.git
+$ source poky/oe-init-build-env
+$ bitbake-layers add-layer ../meta-openembedded/meta-oe ../meta-clang ../meta-python2 ../meta-browser/meta-chromium
+```
+And, append target as I in conf/local.conf.
+
+```sh
+IMAGE_INSTALL_append = " chromium-ozone-wayland "
+```
+
+Then, execute bitbake command to build an image.
+
 Contributing
 ------------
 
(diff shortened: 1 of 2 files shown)


meta-chromium/README.md
```diff
@@ -23,6 +23,50 @@ This layer depends on:
...
+
+BBLAYERS ?= " \
+  ${TOPDIR}/../poky/meta \
+  ${TOPDIR}/../poky/meta-poky \
+  ${TOPDIR}/../poky/meta-yocto-bsp \
+  ${TOPDIR}/../meta-openembedded/meta-oe \
+  ${TOPDIR}/../meta-clang \
+  ${TOPDIR}/../meta-python2 \
+  ${TOPDIR}/../meta-browser/meta-chromium \
+  "
```
otavio, line 59: You can use `bitbake-layers` tool to do this without manually editing it.

otavio (changes requested):

meta-chromium/README.md
kenhys: Thanks, I've fixed to use bitbake-layers.

otavio (approved):

meta-chromium/README.md
```diff
@@ -23,6 +23,28 @@ This layer depends on:
...
+```sh
+$ git clone git://git.yoctoproject.org/poky
+$ git clone git://git.openembedded.org/meta-openembedded
+$ git clone git://github.com/kraj/meta-clang
+$ git clone git://git.openembedded.org/meta-python2
+$ git clone git://github.com/OSSystems/meta-browser.git
+$ source poky/oe-init-build-env
+$ bitbake-layers add-layer ../meta-openembedded/meta-oe ../meta-clang ../meta-python2 ../meta-browser/meta-chromium
+```
+And, append target as I in conf/local.conf.
```
rakuco, line 40: Sorry, I don't understand this sentence. What target are you appending where, and what does "as I" mean?

meta-chromium/README.md
```diff
@@ -23,6 +23,28 @@ This layer depends on:
   - branch: master
   - revision: HEAD
 
+Usage
+-----
+
+The following steps are typical use case which builds chromium-ozone-wayland for example.
```
rakuco, line 29: I think the most important part of the process for chromium-ozone-wayland is to set things like `DISTRO_FEATURES` appropriately so that Wayland's used instead of X11, but this is not mentioned here.

rakuco (commented):
The instructions look mostly OK, but I'm still wondering if they are really necessary: isn't the process basically identical to any other layer, and thus already described in the OE or Yocto documentation?

meta-chromium/README.md
kenhys: Oops, I've removed garbage.

shr-project:
> The instructions look mostly OK, but I'm still wondering if they are really necessary: isn't the process basically identical to any other layer, and thus already described in the OE or Yocto documentation?

Agreed, also why should meta-browser recommend to use poky instead of checking out oe-core + bitbake separately.

The list of necessary layers is already included in LAYERDEPENDS (which shows reasonable error message when some are missing in your bblayers.conf) and "This layer depends on:" section above. Showing how to use git to clone them IMHO doesn't belong in meta-browser.

meta-chromium/README.md
```diff
@@ -23,6 +23,28 @@ This layer depends on:
...
+$ git clone git://git.openembedded.org/meta-python2
+$ git clone git://github.com/OSSystems/meta-browser.git
+$ source poky/oe-init-build-env
+$ bitbake-layers add-layer ../meta-openembedded/meta-oe ../meta-clang ../meta-python2 ../meta-browser/meta-chromium
+```
+And, append target in conf/local.conf.
+
+```sh
+IMAGE_INSTALL_append = " chromium-ozone-wayland "
+```
```
shr-project, line 44: This "typical" case will also cause chromium-ozone-wayland to be installed in various initramfs* images if you platform uses one, which is definitely not something you want.

Also trailing space isn't needed with _append, only the leading one is.

rakuco:
@kenhys Once again, thank you for writing the patch and submitting it. I think the consensus is that it is not necessary though, given it duplicates instructions available in existing documentation or which is output by the tools we have available, as explained above. Please feel free to reopen if you'd like to discuss it further.

kenhys:
> I think the consensus is that it is not necessary though, 

it seems that it is reasonable.

PR #520, opened 2021-06-21
chromium: fix various build errors

### LINKED ISSUE #517: Chromium build errors
Hello,
Checking out from master and building Chromium currently results in errors.

The compile process fails due to the following error:

```
| ../../third_party/harfbuzz-ng/src/src/hb-subset-cff1.cc:405:33: error: variable 'supp_size' set but not used [-Werror,-Wunused-but-set-variable]
|     unsigned int  size0, size1, supp_size;
|                                 ^
| 1 error generated.
```

This is due to the harfbuzz third-party library, and was fixed in harfbuzz with commit `243d056ff1c2af583ceb67e5dfbfaac51dc96e63`

I created a patch to cherry pick the commit to fix this issue, and updated the chromium-gn.inc file to apply the patch.
This fixed that original error but let to a different one:

```
| ../../third_party/ruy/src/ruy/block_map.cc:375:25: error: no member named 'numeric_limits' in namespace 'std'
|   int best_score = std::numeric_limits<int>::min();
|                    ~~~~~^
| ../../third_party/ruy/src/ruy/block_map.cc:375:43: error: expected '(' for function-style cast or type construction
|   int best_score = std::numeric_limits<int>::min();
|                                        ~~~^
| ../../third_party/ruy/src/ruy/block_map.cc:375:46: error: no member named 'min' in the global namespace
|   int best_score = std::numeric_limits<int>::min();
|                                            ~~^
| 3 errors generated.
```

Similarly, this is due to google's ruy library. There is commit upstream that fixes this: `3c93cda8211efa01128d48950f0d6ee5233c5b9b`, so I created a similar patch for it and applied it.

This fixed this issue but led a final one:

```
| In file included from ../../sql/recover_module/cursor.cc:5:
| ../../sql/recover_module/cursor.h:65:5: error: static_assert failed due to requirement 'std::is_standard_layout<sql::recover::VirtualCursor>::value' "needed for the reinterpret_cast below"
|     static_assert(std::is_standard_layout<VirtualCursor>::value,
|     ^             ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
| 1 error generated.
```
Unlike the others, this is not fixed upstream yet.
We're led to believe this may be due to stricter compiler rules in master for oe now.

I have attached the 2 patches to fix the earlier issues as reference.

Thanks!

[fix-harfbuzz-supp-size.patch.txt](https://github.com/OSSystems/meta-browser/files/6673383/fix-harfbuzz-supp-size.patch.txt)
[fix-ruy-numeric-limits.patch.txt](https://github.com/OSSystems/meta-browser/files/6673385/fix-ruy-numeric-limits.patch.txt)

TonyTascioglu:
Fix several errors during build due to stricter clang builds.
  - unused variable error in harfbuzz third party dependency
  - missing include in ruy third party dependency
  - virtualcursor error in sql/recover_module/cursor.cc

The third party fixes are back-ported from upstream.
The virtualcursor patch is from ungoogled-chromium.

Fixes #517

Signed-off-by: Tony Tascioglu <tony.tascioglu@windriver.com>

### DIFF
--- a/meta-chromium/recipes-browser/chromium/files/fix-harfbuzz-supp-size-error.patch
+++ b/meta-chromium/recipes-browser/chromium/files/fix-harfbuzz-supp-size-error.patch
@@ -1,0 +1,49 @@
+From 243d056ff1c2af583ceb67e5dfbfaac51dc96e63 Mon Sep 17 00:00:00 2001
+From: Andi-Bogdan Postelnicu <abpostelnicu@me.com>
+Date: Wed, 2 Jun 2021 14:08:11 +0300
+Subject: [PATCH] Removed unused variable `supp_size` from
+ plan_subset_encoding(...).
+
+Builds with the latest layers fail as a result of stricter compiler
+checks. This patch backports a fix upstream to remove an unused variable
+which casues the build error. 
+
+Upstream-Status: Backport [https://github.com/harfbuzz/harfbuzz/commit/243d056ff1c2af583ceb67e5dfbfaac51dc96e63]
+
+Signed-off-by: Tony Tascioglu <tony.tascioglu@windriver.com>
+---
+ third_patry/harfbuzz-ng/src/src/hb-subset-cff1.cc | 4 +---
+ 1 file changed, 1 insertion(+), 3 deletions(-)
+
+diff --git a/third_party/harfbuzz-ng/src/src/hb-subset-cff1.cc b/third_party/harfbuzz-ng/src/src/hb-subset-cff1.cc
+index df322f845..35dae7b1f 100644
+--- a/third_party/harfbuzz-ng/src/src/hb-subset-cff1.cc
++++ b/third_party/harfbuzz-ng/src/src/hb-subset-cff1.cc
+@@ -402,7 +402,7 @@ struct cff_subset_plan {
+   void plan_subset_encoding (const OT::cff1::accelerator_subset_t &acc, hb_subset_plan_t *plan)
+   {
+     const Encoding *encoding = acc.encoding;
+-    unsigned int  size0, size1, supp_size;
++    unsigned int  size0, size1;
+     hb_codepoint_t  code, last_code = CFF_UNDEF_CODE;
+     hb_vector_t<hb_codepoint_t> supp_codes;
+ 
+@@ -412,7 +412,6 @@ struct cff_subset_plan {
+       return;
+     }
+ 
+-    supp_size = 0;
+     supp_codes.init ();
+ 
+     subset_enc_num_codes = plan->num_output_glyphs () - 1;
+@@ -448,7 +447,6 @@ struct cff_subset_plan {
+ 	  code_pair_t pair = { supp_codes[i], sid };
+ 	  subset_enc_supp_codes.push (pair);
+ 	}
+-	supp_size += SuppEncoding::static_size * supp_codes.length;
+       }
+     }
+     supp_codes.fini ();
+-- 
+2.32.0
+
--- a/meta-chromium/recipes-browser/chromium/files/fix-ruy-numeric-limits-error.patch
+++ b/meta-chromium/recipes-browser/chromium/files/fix-ruy-numeric-limits-error.patch
@@ -1,0 +1,31 @@
+From 3c93cda8211efa01128d48950f0d6ee5233c5b9b Mon Sep 17 00:00:00 2001
+From: stha09 <51720730+stha09@users.noreply.github.com>
+Date: Thu, 6 May 2021 18:31:30 +0200
+Subject: [PATCH] IWYU: include limits for std::numeric_limits (#253)
+
+Builds with the latest layers fail as a result of stricter compiler
+checks. This patch backports a fix for a missing include statement in
+the third party dependency on ruy.
+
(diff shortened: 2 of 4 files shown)


kraj (approved):
Lgtm

msisov:
Can you please specify what yocto branch and toolchains require these changes?

meta-chromium/recipes-browser/chromium/files/fix-harfbuzz-supp-size-error.patch
```diff
@@ -0,0 +1,43 @@
+From 243d056ff1c2af583ceb67e5dfbfaac51dc96e63 Mon Sep 17 00:00:00 2001
+From: Andi-Bogdan Postelnicu <abpostelnicu@me.com>
+Date: Wed, 2 Jun 2021 14:08:11 +0300
+Subject: [PATCH] Removed unused variable `supp_size` from
+ plan_subset_encoding(...).
+
+Upstream-Status: Backport [https://github.com/harfbuzz/harfbuzz/commit/243d056ff1c2af583ceb67e5dfbfaac51dc96e63]
```
msisov, line 7: This should be the first 6 lines .. eg -

Upstream-Status: Backport

Comment + link

Signed-off-by: Name <email@email.com>
----

meta-chromium/recipes-browser/chromium/files/fix-ruy-numeric-limits-error.patch
```diff
@@ -0,0 +1,25 @@
+From 3c93cda8211efa01128d48950f0d6ee5233c5b9b Mon Sep 17 00:00:00 2001
+From: stha09 <51720730+stha09@users.noreply.github.com>
+Date: Thu, 6 May 2021 18:31:30 +0200
+Subject: [PATCH] IWYU: include limits for std::numeric_limits (#253)
+
+Upstream-Status: Backport [https://github.com/google/ruy/commit/3c93cda8211efa01128d48950f0d6ee5233c5b9b]
```
msisov, line 6: ditto

meta-chromium/recipes-browser/chromium/files/fix-sql-virtualcursor-error.patch
```diff
@@ -0,0 +1,233 @@
...
+sql::recover::VirtualCursor needs to be a standard layout type, but
+has members of type std::unique_ptr. However, std::unique_ptr is not
+guaranteed to be standard layout. Compiling with clang combined with
+gcc-11 libstdc++ fails because of this. Replace std::unique_ptr with
+raw pointers.
+
+Bug: 1189788
+Change-Id: Ia6dc388cc5ef1c0f2afc75f8ca45b9f12687ca9c
+
+Upstream-Status: Pending [https://github.com/ungoogled-software/ungoogled-chromium-archlinux/blob/master/sql-VirtualCursor-standard-layout.patch]
```
msisov, line 15: ditto.

rakuco (commented):
Thanks for the patches and for submitting your first PR to this repository!

Before merging, I'd like to wait for some answers to the questions I asked in #517. If this is really due to native builds depending on a libstdc++ that is too old, it'd be good to mention this in the patches so that they can be easily removed if/when we manage to build everything with libc++ in the future. Before #492 it was really annoying to have to chase/write/carry multiple patches to support libstdc++ in the tree so I'd like to keep that to a minimum.

meta-chromium/recipes-browser/chromium/files/fix-harfbuzz-supp-size-error.patch
TonyTascioglu: Sorry if I'm misunderstanding, am I just missing the Signed-off-by lines in the patches, or also a comment for upstream status? Would the comment be an explanation for why the patch is required?

kraj: yes. you can look for patch header template here https://www.openembedded.org/wiki/Commit_Patch_Message_Guidelines

rakuco (commented):
Sorry for the delay here. I think the Harfbuzz patch definitely makes sense, but I'm wondering about the Ruy and SQL ones.

If I understood things correctly, they are all needed because of build failures with libstdc++, which are caused by the native side of the build. Harfbuzz is definitely used by some native dependencies, but I can't see that with Ruy or `//sql`.

I'm on hardknott here, with the latest meta-browser master, and I've run the following commands inside devshell:

```
gn refs out/Release '//sql:sql(//build/toolchain/yocto:yocto_native)'
gn refs out/Release '//third_party/ruy:ruy(//build/toolchain/yocto:yocto_native)'
```

and did not get any matches.

Can you just confirm if these are all failures you're getting in host builds? The other possibility at least for Ruy is that this affects libc++/clang from meta-clang master, but this still wouldn't explain the SQL patch, as it seems to fix a very libstdc++-specific issue.

schnitzeltony:
What happened to this one? Have it in my builds to fix breaks with latest meta-clang

kraj:
I have bundled this pull into https://github.com/OSSystems/meta-browser/pull/529

otavio:
@TonyTascioglu as @kraj has bundled it with #529, I am closing this PR and any further discussion about it should be moved there.

PR #532, opened 2021-08-02
layer.conf: Add honister to LAYERSERIES_COMPAT

ckalluri73:
Add hoister to LAYERSERIES_COMPAT

Signed-off-by: Sai Hari Chandana Kalluri <chandana.kalluri@xilinx.com>

### DIFF
--- a/meta-chromium/conf/layer.conf
+++ b/meta-chromium/conf/layer.conf
@@ -9,6 +9,6 @@ BBFILE_PATTERN_chromium-browser-layer := "^${LAYERDIR}/"
 BBFILE_PRIORITY_chromium-browser-layer = "7"
 
 LAYERVERSION_chromium-browser-layer = "1"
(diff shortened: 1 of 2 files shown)


shr-project:
There is no point in adding it without changing the overrides syntax as done in https://github.com/OSSystems/meta-browser/pull/530

kraj:
> There is no point in adding it without changing the overrides syntax as done in #530

I think it will be needed regardless if override change or not towards the release. So perhaps it can be applied independent of override changes.

shr-project:
It doesn't parse with current honister, marking it as compatible with honister without override changes to make it compatible with honister is wrong.

otavio:
I agree with @shr-project here. We ought to put it as part of #530 so we handle it together.

@ckalluri73 as this is already included in #530 I'll close this PR.

PR #534, opened 2021-08-12
Add missing dependency on curl and fix build with musl

kraj:


otavio (approved):

kraj:
Please merge this since it will be needed for building with glibc < 2.34 ( 2.34 is in master-next but not yet in master )

Issue #541, opened 2021-08-18
Chromium does not run due to missing crashpad handler?


MastaG:
Hi there,

After updating to 92.0.4515.107 I can't seem to run Chromium any longer.

[21439:21439:0818/092005.587570:FATAL:double_fork_and_exec.cc(131)] execv /usr/lib/chromium/crashpad_handler: No such file or directory (2)

The binary doesn't seem to exist for me.
Any ideas?

kraj:
It is built but not installed perhaps you can easily cook a patch to add an install command in do_install

MastaG:
Yes, I'll push a PR soon.

barndoor2276:
I was running into this issue this week also. I installed the binary via a do_install_append in a bbappend.
```
do_install_append() {
	install -m 0755 crashpad_handler ${D}${libdir}/chromium/
}
```
That worked but chromium still won't run and a new error pops up. Not sure if the error is related to crashpad or if it's just another error that was being masked.

`[0819/155701.539273:FATAL:semaphore_posix.cc(99)] Check failed: . sem_timedwait: Invalid argument (22)`

kraj:
@barndoor2276 which architecture/machine was targeted

barndoor2276:
@kraj 
arm
imx6q-dms-ba16

MastaG:
I'm on hardknott and it's working fine for me on armv7 (hard float).
Perhaps the crashpad_handler isn't built for each target (because not everyone is complaining), so I think we should guard it with some if-statement.
"""
if [ -e crashpad_handler ]
then
    install -m 0755 crashpad_handler ${D}${libdir}/chromium/
""""

kraj:
> I'm on hardknott and it's working fine for me on armv7 (hard float).
> Perhaps the crashpad_handler isn't built for each target (because not everyone is complaining), so I think we should guard it with some if-statement.
> """
> if [ -e crashpad_handler ]
> then
> install -m 0755 crashpad_handler ${D}${libdir}/chromium/
> """"

I think newer chromiums should all build it but its fine to see if the binary is existing before packaging it just to make robust.

kraj:
> I was running into this issue this week also. I installed the binary via a do_install_append in a bbappend.
> 
> ```
> do_install_append() {
> 	install -m 0755 crashpad_handler ${D}${libdir}/chromium/
> }
> ```
> 
[...]

PR #552, opened 2021-09-01
chromium: Package crashpad_handler

kraj:
Fixes
https://github.com/OSSystems/meta-browser/issues/541

### DIFF
--- a/meta-chromium/recipes-browser/chromium/chromium-gn.inc
+++ b/meta-chromium/recipes-browser/chromium/chromium-gn.inc
@@ -469,6 +469,9 @@ do_install() {
 
 	# ChromeDriver.
 	install -m 0755 chromedriver.unstripped ${D}${bindir}/chromedriver
+
+	# Install crashpad_handler, it is renamed in newer versions so be aware when upgrading to M94+
+	install -m 0755 crashpad_handler ${D}${libdir}/chromium/
 }
 
 PACKAGES =+ "${PN}-chromedriver"


meta-chromium/recipes-browser/chromium/chromium-gn.inc
```diff
@@ -469,6 +469,10 @@ do_install() {
 
 	# ChromeDriver.
 	install -m 0755 chromedriver.unstripped ${D}${bindir}/chromedriver
+        # Install crashpad_handler if built
```
rakuco, line 472: * The indentation in this line looks wrong.
* It'd be good to add an extra empty line between this one and the ChromeDriver block.
* Can you also explain in the comment when this is _not_ built, like we do with the other checks?

kraj: it builts for me, so check is for robustness

rakuco: If it's always supposed to be built, the check could end up masking an issue in the future. I'd like to understand whether it's mandatory or not, and avoid the check if it is.

kraj: its not mandatory as of now.

rakuco: Given https://bugs.chromium.org/p/chromium/issues/detail?id=1176772 I'm pretty sure it's mandatory since M92. And https://bugs.chromium.org/p/chromium/issues/detail?id=1233008#c4 supports the idea of always packaging it, as the binary has a new name as of M94 and this will help us remember we need to modify the recipe once we move to it.

MastaG:
Thanks @kraj 

Sorry I didn't open a MR for this earlier.
I was really busy.

But this exactly how I have it in my local branch as well.

meta-chromium/recipes-browser/chromium/chromium-gn.inc
kraj: OK let me drop the check then

rakuco (approved):
lgtm, thank you.

PR #553, opened 2021-09-02
Firefox: update for Honister

shr-project:


kraj (approved):
LGTM, perhaps send the two patches I marked upstream as well.

meta-firefox/conf/layer.conf
```diff
@@ -115,4 +115,4 @@ BBFILE_PRIORITY_firefox-browser-layer = "7"
 LAYERVERSION_firefox-browser-layer = "1"
 LAYERSERIES_COMPAT_firefox-browser-layer = "dunfell gatesgarth hardknott honister"
```
kraj, line 116: we would need to drop compatibility with older than honister here if we drop rust layer.

shr-project: It's still compatible as long as user adds meta-rust after seeing the error about cargo.bbclass missing. The rest of the changes should be backwards compatible all the way to dunfell.

shr-project:
> LGTM, perhaps send the two patches I marked upstream as well.

This version of firefox 68 is so old that it's most likely already fixed in upstream (version 91) as well. I just wanted to fix it to re-run test-oe-build-time with honister.

Anyone actually using firefox should work on upgrading it first.

PR #551, opened 2021-09-01
Add patch to fix build in latest oe/gcc11 environment

schnitzeltony:
* This is second part of https://github.com/OSSystems/meta-browser/pull/542
* Still necessary for me ti fix build
* Build tested with 92.0.4515.159 / Raspi4 / 64bit

PR #554, opened 2021-09-02
chromium: Set LIBCPLUSPLUS to use libc++

kraj:
This should fix issue seen with
https://github.com/OSSystems/meta-browser/pull/551

### DIFF
--- a/meta-chromium/recipes-browser/chromium/chromium.inc
+++ b/meta-chromium/recipes-browser/chromium/chromium.inc
@@ -34,10 +34,6 @@ RUNTIME = "llvm"
 # This also ensures that it does not rely too much on meta-clang's defaults.
 LIBCPLUSPLUS = "-stdlib=libc++"
 
-# LIBCPLUSPLUS is included in CXXFLAGS and LDFLAGS, but chromium doesn't respect it
-# add it to TUNE_CCARGS as well to include it in CC/CXX variables which are respected.
-TUNE_CCARGS += "${LIBCPLUSPLUS}"
-
 inherit mime-xdg pythonnative
 DEPENDS += "python-setuptools-native"
 
--- a/meta-chromium/recipes-browser/chromium/chromium.inc
+++ b/meta-chromium/recipes-browser/chromium/chromium.inc
@@ -24,6 +24,20 @@ TOOLCHAIN:class-native = "clang"
 # https://github.com/kraj/meta-clang/issues/449).
 RUNTIME = "llvm"
 
+# meta-clang uses RUNTIME variable to add build dependencies on libcxx/compiler-rt
+# but does not add compiler options to commandline. It expects compiler's own
+# defaults to be used and compiler's buildtime defaults are chosen based on
+# global RUNTIME variable setting when building clang compiler, which defaults to 'gnu'.
+# For chromium libc++ is expected to be used therefore ensure that
+# right build flags are set for chromium irrespective of
+# distro defaults (which could be 'gnu' or 'llvm').
+# This also ensures that it does not rely too much on meta-clang's defaults.
+LIBCPLUSPLUS = "-stdlib=libc++"
+
+# LIBCPLUSPLUS is included in CXXFLAGS and LDFLAGS, but chromium doesn't respect it
+# add it to TUNE_CCARGS as well to include it in CC/CXX variables which are respected.
+TUNE_CCARGS += "${LIBCPLUSPLUS}"
+
 inherit mime-xdg pythonnative
 DEPENDS += "python-setuptools-native"
 
--- a/meta-chromium/recipes-browser/chromium/gn-utils.inc
+++ b/meta-chromium/recipes-browser/chromium/gn-utils.inc
@@ -93,11 +93,11 @@ def write_toolchain_file(d, file_path):
         'cxx': d.expand('${CXX}'),
         'ar': d.expand('${AR}'),
         'nm': d.expand('${NM}'),
-        'readelf': d.expand('${TARGET_PREFIX}readelf'),
-        'extra_cflags': d.expand('${TARGET_CFLAGS}'),
-        'extra_cppflags': d.expand('${TARGET_CPPFLAGS}'),
-        'extra_cxxflags': d.expand('${TARGET_CXXFLAGS}'),
-        'extra_ldflags': d.expand('${TARGET_LDFLAGS}'),
+        'readelf': d.expand('${READELF}'),
+        'extra_cflags': d.expand('${CFLAGS}'),
+        'extra_cppflags': d.expand('${CPPFLAGS}'),
+        'extra_cxxflags': d.expand('${CXXFLAGS}'),
+        'extra_ldflags': d.expand('${LDFLAGS}'),
     }
 
     with open(file_path, 'w') as toolchain_file:


rakuco (commented):
Does this also work with older branches like hardknott? In case of master, where we didn't seem to be passing any toolchain-related arguments per #551, doesn't it mean you also need to pass `-rtlib` and `--unwindlib`?

rakuco:
cc @shr-project @schnitzeltony

kraj:
> Does this also work with older branches like hardknott? In case of master, where we didn't seem to be passing any toolchain-related arguments per #551, doesn't it mean you also need to pass `-rtlib` and `--unwindlib`?

yes it will be a nop on hardknott and dunfell.

rakuco:
> In case of master, where we didn't seem to be passing any toolchain-related arguments per #551, doesn't it mean you also need to pass `-rtlib` and `--unwindlib`?

How about this part?

shr-project:
This didn't work for me, because chromium doesn't seem respect CXXFLAGS/LDFLAGS ("stdlib=libc++" doesn't appear anywhere in log.do_compile - even for other modules than sql).

If I add LIBCPLUSPLUS to TUNE_CCARGS (sort of like it was in meta-clang/hardknott), then it builds OK.

meta-chromium/recipes-browser/chromium/chromium.inc
```diff
@@ -24,6 +24,13 @@ TOOLCHAIN:class-native = "clang"
 # https://github.com/kraj/meta-clang/issues/449).
 RUNTIME = "llvm"
 
+# meta-clang uses RUNTIME variable to add dependencies but does not add needed
+# compiler options since it expects compiler buildtime defaults to be chosen
+# based on RUNTIME variable which defaults to 'gnu' which means we have to
+# ensure right cxxflags and ldflags are set too for chromium irrespective of
+# disto defaults which could be 'gnu' or 'llvm'
+LIBCPLUSPLUS = "-stdlib=libc++"
```
schnitzeltony, line 32: Sorry but from comment I don't understand what's going on.
Isn't it much simpler:
RUNTIME takes care of additional DEPENDS but does not touch cxx-/ldflags
=> To force libc++ we have to set LIBCPLUSPLUS

shr-project: I misunderstood it first as well, but I think Khem talks about the default RUNTIME the cross toolchain defaults to (not as RUNTIME oe variable). Because RUNTIME is already set to llvm in chromium-x11, but that doesn't help with this issue as it doesn't affect anything since https://github.com/kraj/meta-clang/commit/dc5e04e5819a5e7384ae48f1740ac82970a80b8f and https://github.com/kraj/meta-clang/commit/4d066667c3cbba3b88d2f8d6c5640f5f793d10b4 which removed LIBCPLUSPLUS from TUNE_CCARGS as well.

kraj: yeah most packages do respect CPPFLAGS/LDFLAGS but here in chromium we might be at odds. Perhaps adding `-stlib=xxx` option can be added to both link and compile stages without issues so perhaps we are ok there to use is via TUNE_CCARGS, let me fix the patch to use TUNE_CCARGS here and also improve the comment a bit

kraj: > Sorry but from comment I don't understand what's going on.
> Isn't it much simpler:
> RUNTIME takes care of additional DEPENDS but does not touch cxx-/ldflags
> => To force libc++ we have to set LIBCPLUSPLUS

its not easily possible to allow switching runtime at recipe level, so I would like folks to use it at distro level unless
packages do the right thing to use one or other like chromium, thats why the change is done in meta-clang because in many recipes you may choose LLVM runtime but then it might bring in gnu runtime into build via other dependencies and conflict, then we will end up fixing them which is unwanted work.

shr-project:
Last version works for me +1

schnitzeltony:
Builds fine - thanks

meta-chromium/recipes-browser/chromium/chromium.inc
```diff
@@ -24,6 +24,22 @@ TOOLCHAIN:class-native = "clang"
...
+# global RUNTIME variable setting, which defaults to 'gnu'
+# For chromium libc++ is expected to be used so, irrespective of
+# ensure right build flags are set too for chromium irrespective of
+# disto defaults which could be 'gnu' or 'llvm', here we need to set both
+# target and buid flags since chromium builds native pieces along with target
+# pieces and does not have separate native variant to provide the native tools
+# this also ensures that it does not rely too much on meta-clang's defaults
+LIBCPLUSPLUS = "-stdlib=libc++"
+
+# LIBCPLUSPLUS is included in CXXFLAGS and LDFLAGS, but chromium doesn't respect it
```
rakuco, line 39: Hmm, does it really not respect those variables? The logic in [gn-utils.inc's write_toolchain_file()](https://github.com/OSSystems/meta-browser/blob/3405e3af0fa67a82bd28884cac47314bf847e6b7/meta-chromium/recipes-browser/chromium/gn-utils.inc#L49) reads all the {BUILD,TARGET}_{CPP,CXX,LD,CC}FLAGS, so I'm wondering if there's something to fix there instead.

shr-project: CXXFLAGS != TARGET_CXXFLAGS (LIBCPLUSPLUS is appended to CXXFLAGS but not TARGET_CXXFLAGS).

```
# $CXXFLAGS [7 operations]                                                                                                                                                   |
#   :append[pn-xdelta3:toolchain-clang] /jenkins/mjansa/build/nodistro/honister/meta-clang/conf/nonclangable.conf:273                                                        |
#     " -Wno-error=reserved-user-defined-literal"                                                                                                                            |
#   :remove[pn-mozjs:toolchain-clang] /jenkins/mjansa/build/nodistro/honister/meta-clang/conf/nonclangable.conf:275                                                          |
#     "-fno-tree-vrp"                                                                                                                                                        |
#   :append[pn-apt:toolchain-clang] /jenkins/mjansa/build/nodistro/honister/meta-clang/conf/nonclangable.conf:353                                                            |
#     " -Wno-c++11-narrowing"                                                                                                                                                |
#   exported /jenkins/mjansa/build/nodistro/honister/oe-core/meta/conf/bitbake.conf:579                                                                                      |
#     [export] "1"                                                                                                                                                           |
#   set /jenkins/mjansa/build/nodistro/honister/oe-core/meta/conf/bitbake.conf:579                                                                                           |
#     "${TARGET_CXXFLAGS}"                                                                                                                                                   |
#   append /jenkins/mjansa/build/nodistro/honister/oe-core/meta/conf/distro/include/tclibc-glibc.inc:17                                                                      |
#     "-fvisibility-inlines-hidden"                                                                                                                                          |
#   :append[toolchain-clang] /jenkins/mjansa/build/nodistro/honister/meta-clang/classes/clang.bbclass:29                                                                     |
#     " ${LIBCPLUSPLUS}"                                                                                                                                                     |
# pre-expansion value:                                                                                                                                                       |
#   "${TARGET_CXXFLAGS} -fvisibility-inlines-hidden ${LIBCPLUSPLUS}"                                                                                                         |
export CXXFLAGS=" -O2 -pipe  -feliminate-unused-debug-types -fmacro-prefix-map=/jenkins/mjansa/build/nodistro/honister/BUILD/work/core2-32-oe-linux/chromium-x11/92.0.4515.15|
9-r0=/usr/src/debug/chromium-x11/92.0.4515.159-r0                      -fdebug-prefix-map=/jenkins/mjansa/build/nodistro/honister/BUILD/work/core2-32-oe-linux/chromium-x11/9|
2.0.4515.159-r0=/usr/src/debug/chromium-x11/92.0.4515.159-r0                      -fdebug-prefix-map=/jenkins/mjansa/build/nodistro/honister/BUILD/work/core2-32-oe-linux/chr|
omium-x11/92.0.4515.159-r0/recipe-sysroot=                      -fdebug-prefix-map=/jenkins/mjansa/build/nodistro/honister/BUILD/work/core2-32-oe-linux/chromium-x11/92.0.451|
5.159-r0/recipe-sysroot-native= -g1 -fvisibility-inlines-hidden -stdlib=libc++"  
```

Changing:
```
meta-chromium/recipes-browser/chromium/gn-utils.inc:        'extra_cxxflags': d.expand('${TARGET_CXXFLAGS}'),
```
To CXXFLAGS should work.

rakuco: > CXXFLAGS != TARGET_CXXFLAGS (LIBCPLUSPLUS is appended to CXXFLAGS but not TARGET_CXXFLAGS).

Ah, I see, thanks for the clarification. I wonder if LIBCPLUSPLUS shouldn't be added to TARGET_CXXFLAGS instead? According to the docs CXXFLAGS is set as an environment variable in the compilation step; I think if it's set during the configuration step it should work with GN as well, but if we can capture all necessary flags when write_toolchain_file() is set it's even better to avoid discrepancies and flags showing up out of nowhere.

shr-project: Using variables without TARGET_ prefix is correct and e.g. cmake is using the same when generating its toolchain config:
https://git.openembedded.org/openembedded-core/tree/meta/classes/cmake.bbclass#n28

I've verified that LIBCPLUSPLUS works with: https://github.com/shr-project/meta-browser/commit/1525aa425df2fce4ad7a98f8b3e2b8716fff6b08

shr-project:
LGTM

meta-chromium/recipes-browser/chromium/chromium.inc
```diff
@@ -24,6 +24,18 @@ TOOLCHAIN:class-native = "clang"
 # https://github.com/kraj/meta-clang/issues/449).
 RUNTIME = "llvm"
 
+# meta-clang uses RUNTIME variable to add build dependencies on libcxx/compiler-rt
+# but does not add compiler options to commandline since it expects compiler's own
+# defaults to be used and compiler's buildtime defaults are chosen based on
+# global RUNTIME variable setting, which defaults to 'gnu'
+# For chromium libc++ is expected to be used so, irrespective of
```
rakuco, line 31: The first ", irrespective of" seems like a leftover from a previous version of this comment.

meta-chromium/recipes-browser/chromium/chromium.inc
```diff
@@ -24,6 +24,18 @@ TOOLCHAIN:class-native = "clang"
...
+# ensure right build flags are set too for chromium irrespective of
+# disto defaults which could be 'gnu' or 'llvm', here we need to set both
```
rakuco, line 33: disto -> distro
I'd also add a full stop before "here", otherwise you end up with a very long paragraph with only one sentence that's hard to follow.

meta-chromium/recipes-browser/chromium/chromium.inc
```diff
@@ -24,6 +24,18 @@ TOOLCHAIN:class-native = "clang"
...
+# target and buid flags since chromium builds native pieces along with target
```
rakuco, line 34: buid -> build

rakuco (commented):
This looks a lot better, nice!

One thing: do we still need to continue setting `RUNTIME = "llvm"`? I guess we need it at least for older branches, but I'm not sure.

rakuco (commented):
A few remaining questions:
* Do we still need to set `RUNTIME="llvm"`?
* Is this enough to make the build use compiler-rt and the rest of the LLVM toolchain in terms of arguments passed to the compiler/linker?

kraj:
> A few remaining questions:
> 
> * Do we still need to set `RUNTIME="llvm"`?

yes since it will add needed DEPENDs correctly.

> * Is this enough to make the build use compiler-rt and the rest of the LLVM toolchain in terms of arguments passed to the compiler/linker?

yes what we have is enough for now.

rakuco (approved):
lgtm, thank you both for working on this

rakuco:
@kraj @schnitzeltony I have to revert @kraj's commit, it broke the hardknott build for me. Setting `LIBCPLUSPLUS` actually started preventing the libc++ headers from being installed into `recipe-sysroot/usr/include`.

PR #561, opened 2021-09-14
chromium: Remove unused patch from files/.

rakuco:
`0001-Fixed-build-with-use_ozone-false.patch` stopped being used with commit
157c2dcc26304edc2 ("chromium: update to 92.0.4515.107").


rakuco:
ping @msisov

msisov (approved):

PR #570, opened 2021-09-28
chromium: use bundled freetype

### LINKED ISSUE #569: chromium: Consider using bundled FreeType copy
Chromium project members contribute to FreeType itself, and are always actively tracking the latest upstream versions (include pre-release ones), depending on features included there and staying up-to-date with security issues.

This makes it difficult for us to get Chromium to build on multiple branches, including old ones like dunfell, where the FreeType version might lag behind what Chromium expects by quite a bit.

So far this means we've had commits like ac9d7200001434e692abad4052d5f34ba48f25e9 (with a patch that needs to be updated on every Chromium update), but according to #568 it no longer seems to be enough.

I'd like us to consider passing `use_system_freetype=false` to GN and using Chromium's own FreeType copy.

shr-project:
* Fixes #569

* because the current skia codebase expects freetype newer than latest
  2.11 release from freetype upstream

* chromium-ozone-wayland as well as chromium-x11 with honister based builds:

  ../../third_party/skia/src/ports/SkFontHost_FreeType.cpp:1110:13: error: unknown type name 'FT_ClipBox'
              FT_ClipBox colrGlyphBbox;
              ^

  and

  ../../third_party/skia/src/ports/SkFontHost_FreeType_common.cpp:438:14: error: use of undeclared identifier 'FT_COLR_COMPOSITE_PLUS'
  ../../third_party/skia/src/ports/SkFontHost_FreeType_common.cpp:1009:5: error: unknown type name 'FT_ClipBox'

  FT_COLR_COMPOSITE_PLUS is defined in bundled freetype:
  ./third_party/freetype/src/include/freetype/ftcolor.h:    FT_COLR_COMPOSITE_PLUS           = 12,
  and added to upstream freetype in:
  https://github.com/freetype/freetype/commit/e40ae7569aa4ef591f66ff9066df2f91de75bb77#diff-81347c1a5d4e3ac2906ad0df93738046c8c139135e2b3c1fbfe0ae22f9de929c
  but isn't included in 2.11 version currently in oe-core:
  https://github.com/freetype/freetype/blob/VER-2-11-0/include/freetype/ftcolor.h

  Similarly FT_ClipBox was added after 2.11 in
  https://github.com/freetype/freetype/commit/47cf8ebf4a78ed42da455a98d77a92ce6a180d78#diff-81347c1a5d4e3ac2906ad0df93738046c8c139135e2b3c1fbfe0ae22f9de929c


rakuco:
> * [chromium: Consider using bundled FreeType copy #569](https://github.com/OSSystems/meta-browser/issues/569)
> *

You can say "Fixes #569" in PR and commit messages to automatically close the issue when this is merged.

meta-chromium/recipes-browser/chromium/chromium-gn.inc
```diff
@@ -140,7 +138,6 @@ GN_ARGS = " \
         use_gnome_keyring=false \
         use_kerberos=false \
         use_pulseaudio=${@bb.utils.contains('DISTRO_FEATURES', 'pulseaudio', 'true', 'false', d)} \
-        use_system_freetype=true \
```
rakuco, line 143: Although not required, I'd add a separate block below where we set this to `false` with a comment explaining that we prefer this bundled copy because Chromium targets the latest upstream quite aggressively and depends on features that are not always present in the versions available in the branches we support.

meta-chromium/recipes-browser/chromium/chromium-gn.inc
```diff
@@ -122,6 +120,10 @@ PACKAGECONFIG[proprietary-codecs] = ' \
         ffmpeg_branding="Chrome" proprietary_codecs=true, \
         ffmpeg_branding="Chromium" proprietary_codecs=false \
 '
+# Not enabled by default, because the freetype version from oe-core
```
rakuco, line 123: I meant a block like this one: https://github.com/OSSystems/meta-browser/blob/efd05579551e687c2fe4865c41db0c4fc0f89c70/meta-chromium/recipes-browser/chromium/chromium-gn.inc#L147 rather than a `PACKAGECONFIG` knob. Having a knob for FreeType won't work, as we'll basically never test the bundled option anyway.

shr-project: system freetype works for dunfell and hardknott, right? So having a knob for people who really want to use system freetype (even if they need to provide additional patch) would be more convenient for them than forcing them to use GN_ARGS:remove

kraj: I understand that would cause additional support calls here but I guess I am not too worried so keep it as pkgconfig is good

rakuco: I'd really like to reduce the amount of configurations we need to support. Having this as a `PACKAGECONFIG` knob means we'd have to keep `0012-chromium-freetype-2.11.patch` around and test it on the older branches. If we're only going to test the bundled case, I don't see why keep the non-bundled option if it will quickly bit-rot.

Plus I think the very fact that the older branches built M94 but honister and later did not indicates that the Chromium code is detecting different features based on the branches, which can also lead to unintended rendering bugs. If we're going to use Chromium's bundled FreeType, I think it's much better to do it unconditionally.

kraj: yeah I think we should just leave the option out and use internal freetype

rakuco (approved):
lgtm

PR #572, opened 2021-09-28
chromium: update typo in a freetype patch

msisov:


rakuco (approved):
lgtm, but #570 will drop this patch altogether if we reach an agreement in #569.

msisov:
> lgtm, but #570 will drop this patch altogether if we reach an agreement in #569.

ack.

PR #576, opened 2021-10-10
chromium: add PACKAGECONFIG for gtk

MarkusVolk:
Recently i found out, that chromium needs to be built with use_gtk=true if you want it to respect the gtk cursor theme.
This seems to be related to a change in use flags in chromium. There has been a use_gtk3 option that was true by
default. In addition we had use_gtk to enable builds with gtk+2.

Now it looks like we only have use_gtk to enable build with gtk+
https://github.com/chromium/chromium/blob/72ceeed2ebcd505b8d8205ed7354e862b871995e/build/config/linux/gtk/gtk.gni#L9

We have this disabled by default so i guess all current builds are without using gtk:
https://github.com/OSSystems/meta-browser/blob/8df26e66c28b43f1630fcd2cb952a1e69b6e6558/meta-chromium/recipes-browser/chromium/chromium-ozone-wayland_94.0.4606.71.bb#L30

Make gtk a PACKAGECONFIG to have an option.

Signed-off-by: MarkusVolk <f_l_k@t-online.de>

### DIFF
--- a/meta-chromium/recipes-browser/chromium/chromium-gn.inc
+++ b/meta-chromium/recipes-browser/chromium/chromium-gn.inc
@@ -68,7 +68,6 @@ DEPENDS += " \
     glib-2.0 \
     gn-native \
     gperf-native \
-    gtk+3 \
     jpeg \
     libdrm \
     libwebp \
@@ -105,8 +104,6 @@ BUILD_CC:toolchain-clang = "clang"
 BUILD_CXX:toolchain-clang = "clang++"
 BUILD_LD:toolchain-clang = "clang"
 
-PACKAGECONFIG ??= "upower use-egl"
-
 # this makes sure the dependencies for the EGL mode are present; otherwise, the configure scripts
 # automatically and silently fall back to GLX
 PACKAGECONFIG[use-egl] = ",,virtual/egl virtual/libgles2"
@@ -117,6 +114,7 @@ PACKAGECONFIG[use-egl] = ",,virtual/egl virtual/libgles2"
 # be necessary but are OK to add).
 PACKAGECONFIG[component-build] = ""
 PACKAGECONFIG[cups] = "use_cups=true,use_cups=false,cups"
+PACKAGECONFIG[gtk] = "use_gtk=true,use_gtk=false,gtk+3"
 PACKAGECONFIG[kiosk-mode] = ""
 PACKAGECONFIG[proprietary-codecs] = ' \
         ffmpeg_branding="Chrome" proprietary_codecs=true, \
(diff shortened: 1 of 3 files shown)


meta-chromium/recipes-browser/chromium/chromium-gn.inc
```diff
@@ -117,6 +116,7 @@ PACKAGECONFIG[use-egl] = ",,virtual/egl virtual/libgles2"
 # be necessary but are OK to add).
 PACKAGECONFIG[component-build] = ""
 PACKAGECONFIG[cups] = "use_cups=true,use_cups=false,cups"
+PACKAGECONFIG[gtk] = "use_gtk=true,use_gtk=false,gtk+3"
```
msisov, line 119: gtk used to be true by default on X11 builds.

msisov (changes requested):

meta-chromium/recipes-browser/chromium/chromium-gn.inc
MarkusVolk: I did a test build for chromium-x11 on aarch64 yesterday and explicitly disabled gtk ... it completed without errors.
Seems like meanwhile using gtk is completely optional
Should i make gtk the default again like this?
`PACKAGECONFIG ??= "gtk upower use-egl"`
Should it be default for both, x11 and wayland builds?

msisov: My expectation is that gtk is optional, but it is set by default on X11, but disabled on Wayland.

MarkusVolk: Works like this. args.gn shows 'use_gtk = true' for chromium-x11

MarkusVolk:
my idea was to reflect the fact, that gtk+3 is not a hard dependency anymore and could be left out. This i thought could be useful for non gtk builds. I guess it would be also possible to run chromium under a qt environment ?

A second thing i thought about was, that it might come easier to handle gtk in the bridging time between gtk+3/gtk+4. We would just need to add the PACKAGECONFIG like this:

PACKAGECONFIG[gtk3] = "use_gtk=true,use_gtk=false,gtk+3"
PACKAGECONFIG[gtk4] = "use_gtk=true,use_gtk=false,gtk+4"

Anyway my personal motivation on this is quite low. It's also possible to just pass 'use_gtk=true' in a chromium-ozone-wayland.bbappend to fix the wayland gtk issue.
If you don't want to support building without gtk+ then your solution would be probably the most clean one and would bypass the  problem as well.

MastaG:
Perhaps a bit unrelated to this MR.
But what's the benefit of disabling GTK+ support?
For example isn't it required to get the native file-choser/picker when running it on Gnome Desktop?

Issue #578, opened 2021-10-25
chromium: libsync includes the nonexistant `sys/cdefs.h` in musl


concatime:
Hi, when building Chromium v94.0.4606.61, I get:
```
In file included from ../chromium-94.0.4606.61/third_party/libsync/src/sync.c:30:
../chromium-94.0.4606.61/third_party/libsync/src/include/sync/sync.h:22:10: fatal error: 'sys/cdefs.h' file not found
#include <sys/cdefs.h>
         ^~~~~~~~~~~~~
```
Shouldn’t there be a patch for libsync in the musl directory too ?

Also... thank you for the awesomeness of this project, very helpful!

concatime:
Something like this:
```patch
diff --git a/third_party/libsync/src/include/sync/sync.h b/third_party/libsync/src/include/sync/sync.h
index 50ed0ac5..014ea72d 100644
--- a/third_party/libsync/src/include/sync/sync.h
+++ b/third_party/libsync/src/include/sync/sync.h
@@ -19,12 +19,13 @@
 #ifndef __SYS_CORE_SYNC_H
 #define __SYS_CORE_SYNC_H
 
-#include <sys/cdefs.h>
 #include <stdint.h>
 
 #include <linux/types.h>
 
-__BEGIN_DECLS
+#ifdef __cplusplus
+extern "C" {
+#endif
 
 struct sync_legacy_merge_data {
  int32_t fd2;
@@ -158,6 +159,8 @@ struct sync_pt_info *sync_pt_info(struct sync_fence_info_data *info,
                                   struct sync_pt_info *itr);
 void sync_fence_info_free(struct sync_fence_info_data *info);
 
-__END_DECLS
+#ifdef __cplusplus
+}
+#endif
 
 #endif /* __SYS_CORE_SYNC_H */
```

concatime:
I confirm. With these three patches (nasm, libsync and libevent), I’m able to compile chromium. Would you accept if I submit them, @kraj ?

kraj:
> I confirm. With these three patches (nasm, libsync and libevent) for musl, I’m able (~ish, but that’s for another story) to compile chromium. Would you accept if I submit them, @kraj ?

Yes I will but I am still not sure why
The build is working for me but not for
You I am using arch Linux host and Ubuntu 18.04 hosts to build and it always builds and runs fine for qemux86-64/musl so help us understand how to reproduce the problem

concatime:
I don’t know, but let’s try.

[...]

PR #580, opened 2021-10-30
Fix musl chromium build

concatime:
Fixes https://github.com/OSSystems/meta-browser/issues/578.


kraj (changes requested):
I dont see these patches being applied via chromium-gn.inc so how are they supposed to work?
secondly, I am still bewildered on how to reproduce the failures you are seeing. There much be something thats different in your environment that I am not seeing here. Since build succeeds without these patches for me.

PR #584, opened 2021-11-04
chromium-: Do not use append with += operator

kraj:
this is undefined behaviour, mant times devs used them together to get
the missing space at the beginning of string which append/prepend needs
but thats not intended behaviour


rakuco (approved):
lgtm

PR #586, opened 2021-11-15
chromium: Forward port musl patches to 95.0.4638.69

kraj:

### DIFF
--- a/meta-chromium/recipes-browser/chromium/files/musl/fix-bits-stdint-intn-include.patch
+++ b/meta-chromium/recipes-browser/chromium/files/musl/fix-bits-stdint-intn-include.patch
@@ -1,0 +1,11 @@
+--- a/ui/ozone/platform/wayland/host/wayland_window.cc.orig
++++ b/ui/ozone/platform/wayland/host/wayland_window.cc
+@@ -4,7 +4,7 @@
+ 
+ #include "ui/ozone/platform/wayland/host/wayland_window.h"
+ 
+-#include <bits/stdint-intn.h>
++#include <stdint.h>
+ #include <wayland-cursor.h>
+ #include <algorithm>
+ #include <memory>
--- a/meta-chromium/recipes-browser/chromium/chromium-gn.inc
+++ b/meta-chromium/recipes-browser/chromium/chromium-gn.inc
@@ -50,6 +50,8 @@ SRC_URI:append:libc-musl = "\
     file://musl/0022-sys-stat.patch \
     file://musl/scoped-file.patch \
     file://musl/fix-narrowing-cast.patch \
+    file://musl/chromium-95-quiche-include.patch \
+    file://musl/fix-bits-stdint-intn-include.patch \
 "
 
 ANY_OF_DISTRO_FEATURES = "opengl vulkan"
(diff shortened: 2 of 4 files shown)


meta-chromium/recipes-browser/chromium/files/musl/fix-bits-stdint-intn-include.patch
```diff
@@ -0,0 +1,11 @@
+--- a/ui/ozone/platform/wayland/host/wayland_window.cc.orig
++++ b/ui/ozone/platform/wayland/host/wayland_window.cc
+@@ -4,7 +4,7 @@
+ 
+ #include "ui/ozone/platform/wayland/host/wayland_window.h"
+ 
+-#include <bits/stdint-intn.h>
++#include <stdint.h>
```
rakuco, line 8: @msisov this looks like something that could be upstreamed. It's unusual to include headers in `bits/` rather than their counterparts in `/usr/include` directly. Was there a specific reason for this to happen here?

meta-chromium/recipes-browser/chromium/files/musl/chromium-95-harfbuzz-3.patch
```diff
@@ -0,0 +1,31 @@
+From 9e00e702633c47c590a869bc66b5c2ceec09da50 Mon Sep 17 00:00:00 2001
+From: Dominik Röttsches <drott@chromium.org>
+Date: Tue, 28 Sep 2021 15:31:50 +0000
+Subject: [PATCH] Roll src/third_party/harfbuzz-ng/src/ 6602cbb70..a52c6df38 (80 commits)
```
rakuco, line 4: This looks unrelated to musl. Can you explain what was breaking before?

meta-chromium/recipes-browser/chromium/files/musl/fix-bits-stdint-intn-include.patch
kraj: perhaps yes.

meta-chromium/recipes-browser/chromium/files/musl/chromium-95-harfbuzz-3.patch
kraj: I did not build with glibc so maybe it happens there too but with OE master which harfbuzz 3.1 I get runtime errors e.g.

```
browser log: |
        {"type":"error","text":"Error: Browser exited unexpectedly"}
        {"type":"error","text":"Non-zero exit code: 127"}
        {"type":"error","text":"Error relocating /usr/lib/chromium/chrome: hb_subset_input_set_retain_gids: symbol not found\nError relocating /usr/lib/chromium/chrome: hb_subset: symbol not found\nError relocating /usr/lib/chromium/chrome: hb_subset_input_set_drop_hints: symbol not found\n"}
```

rakuco: I've just finished a chromium-x11 here on honister that did not throw any errors. The version of harfbuzz shipped in OE shouldn't matter, as we build with Chromium's own version (that's often very close to the tip of the tree). Is that not the case for you by any chance?

Also, is the change below enough to stop all those different relocation errors?

kraj: yeah I had external harfbuzz in use, I am testing interrnal one right now if it works then this patch is not needed.

rakuco (approved):
lgtm

meta-chromium/recipes-browser/chromium/files/musl/fix-bits-stdint-intn-include.patch
kraj: submitted https://chromium-review.googlesource.com/c/chromium/src/+/3291865

PR #587, opened 2021-11-20
chromium: enable headless platform again

rwmacleod:
For testing and perhaps other workflows, we should allow headless mode.
This avoids the error:
# chromium --no-sandbox --headless https://html5test.com
[1117/063144.834204:FATAL:platform_selection.cc(45)] Invalid ozone platform: headless
Trace/breakpoint trap


rwmacleod:
Hold on, the default x11 interface might not be the default. More testing needed...

msisov:
This should help https://source.chromium.org/chromium/chromium/src/+/main:build/config/ozone.gni;l=41?q=ozone.gni&ss=chromium :)

PR #593, opened 2021-12-01
chromium-ozone-wayland: fix va-api.

### LINKED ISSUE #592: VA-API patch no longer working
This patch: `0001-ozone-add-va-api-support-to-wayland.patch`

No longer works, as a result of https://chromium-review.googlesource.com/c/chromium/src/+/3141878

However, I can confirm that it does work after adding `properties->supports_vaapi = true;` to `ozone_platform_wayland.cc`, as it is done for X11. `VdaVideoDecoder` is used according to the dev tools and `intel_gpu_top` shows usage of `Video`.

msisov:
After https://crrev.com/c/3141878, it's required to set
supports_vaapi in properties to make Wayland use va-api.

Fixes #592


rakuco (approved):
lgtm

PR #596, opened 2021-12-16
chromium: update to 96.0.4664.110

rwmacleod:
https://chromereleases.googleblog.com/2021/12/stable-channel-update-for-desktop_13.html

The patch, musl/chromium-95-quiche-include.patch was upstreamed by:
   https://quiche.googlesource.com/quiche/+/cde409004ec894d7651a73ba1986ad2efe3683b5%5E%21/#F0
so drop it.

Test-built and run for:
  * chromium-x11
    * master,  clang, MACHINE="qemux86-64,qemuarm64"
    * master,  clang, musl, MACHINE="qemux86-64"

Test-built for:
* chromium-ozone-wayland:
  * honister,  clang, MACHINE="intel-corei7-64"
  * honister,  clang, MACHINE="raspberrypi3
  * hardknott, clang, MACHINE="raspberrypi3
  * dunfell,   clang, MACHINE="qemuarm64"
  * dunfell,   clang, MACHINE="m3ulcb"
* chromium-x11
  * honister,  clang, MACHINE="raspberrypi3"
  * hardknott, clang, MACHINE="raspberrypi3"
  * dunfell,   clang, MACHINE="raspberrypi3"

Signed-off-by: Randy MacLeod <Randy.MacLeod@windriver.com>

PR #597, opened 2021-12-18
chromium-x11: Add missing DEPENDS on libxkbcommon

bobolopolis:
When using a distro without "wayland" in DISTRO_FEATURES, the build
fails due to libxkbcommon being missing. Add the missing libxkbcommon
dependency, since it's required for x11 only builds too.

Signed-off-by: Robert Joslyn <robert.joslyn@redrectangle.org>

### DIFF
--- a/meta-chromium/recipes-browser/chromium/chromium-gn.inc
+++ b/meta-chromium/recipes-browser/chromium/chromium-gn.inc
@@ -72,6 +72,7 @@ DEPENDS += " \
     jpeg \
     libdrm \
     libwebp \
+    libxkbcommon \
     libxslt \
     ninja-native \
     nodejs-native \
--- a/meta-chromium/recipes-browser/chromium/chromium-ozone-wayland_96.0.4664.110.bb
+++ b/meta-chromium/recipes-browser/chromium/chromium-ozone-wayland_96.0.4664.110.bb
@@ -4,7 +4,6 @@ REQUIRED_DISTRO_FEATURES = "wayland"
 
 DEPENDS += "\
         at-spi2-atk \
-        libxkbcommon \
         virtual/egl \
         wayland \
         wayland-native \


rakuco:
Thanks for the patch. What branch/`MACHINE` configuration were you testing? I've finished a chromium-x11 build here on dunfell with MACHINE="qemuarm" that passed, and we tested a lot of different combinations before merging #596, for example.

bobolopolis:
Sorry, I should have put more detail in my commit message. I noticed this on my dunfell builds which use a custom distro and machine, but can reproduce it with just a poky build. I tested this with a stock poky build for qemux86-64 on dunfell. With the default poky DISTRO_FEATURES, chromium-x11 builds fine. If I add `DISTRO_FEATURES_remove = "wayland"` to local.conf, I get this failure:

```| DEBUG: Executing shell function do_configure
| ERROR at //build/config/linux/pkg_config.gni:103:17: Script returned non-zero exit code.
|     pkgresult = exec_script(pkg_config_script, args, "value")
|                 ^----------
| Current dir: /mnt/ssd/yocto/poky/build/tmp/work/core2-64-poky-linux/chromium-x11/96.0.4664.110-r0/chromium-96.0.4664.110/out/Release/
| Command: python3 /mnt/ssd/yocto/poky/build/tmp/work/core2-64-poky-linux/chromium-x11/96.0.4664.110-r0/chromium-96.0.4664.110/build/config/linux/pkg-config.py xkbcommon
| Returned 1.
| stderr:
| 
| Package xkbcommon was not found in the pkg-config search path.
| Perhaps you should add the directory containing `xkbcommon.pc'
| to the PKG_CONFIG_PATH environment variable
| No package 'xkbcommon' found
| Could not run pkg-config.
| 
| See //ui/events/ozone/layout/BUILD.gn:11:3: whence it was called.
|   pkg_config("xkbcommon") {
|   ^------------------------
| See //chrome/test/chromedriver/BUILD.gn:295:7: which caused the file to be included.
|       "//ui/events/ozone/layout",
|       ^-------------------------
```

I assume that the wayland distro feature just causes libxkbcommon to be pulled into the build, so it succeeds on those builds. I haven't tried other branches yet, but I can do so if that would be useful.

rakuco:
> If I add `DISTRO_FEATURES_remove = "wayland"` to local.conf, I get this failure:

I see, I think that's why we weren't running into this issue ourselves, thanks for the explanation.

Could you:
1. Add the error message to the commit message
2. Instead of adding the dependency to the chromium-x11 recipe, remove it from chromium-ozone-wayland and add it to `chromium-gn.inc`?

bobolopolis:
Updated the commit with a more detailed message and moved the dependency to the chromium-gn.inc file.

rakuco (approved):
lgtm, thank you!

PR #598, opened 2021-12-20
chromium-ozone-wayland: fix runtime for wayland 1.20.0

MarkusVolk:
The Ozone/Wayland implementation had a few places where the Wayland
objects were bound without proper checking for their versions.  That was
part of the technical debt not addressed before, and ended up causing
the issue explained in the linked crbug: the compositor terminates the
client that binds to the protocol that it does not actually support.

This patch fixes the issue by adding the necessary checks in all places
where they were missing.  Also a convenience macro for validating the
version is proposed.

author	Alexander Dunaev

Backport patch to fix this issue:
https://bugs.chromium.org/p/chromium/issues/detail?id=1279574

Signed-off-by: MarkusVolk <f_l_k@t-online.de>

### DIFF
--- a/meta-chromium/recipes-browser/chromium/files/chromium-wayland/0001-ozone-wayland-fixed-terminate-caused-by-binding-to-wrong-version.patch
+++ b/meta-chromium/recipes-browser/chromium/files/chromium-wayland/0001-ozone-wayland-fixed-terminate-caused-by-binding-to-wrong-version.patch
@@ -1,0 +1,574 @@
+Upstream-Status: Backport
+
+Fixes "interface 'wl_output' has no event 4" error.
+
+Signed-off-by: Markus Volk f_l_k@t-online.de
+
+author	Alexander Dunaev <adunaev@igalia.com>	Tue Dec 14 09:54:52 2021
+committer	Chromium LUCI CQ <chromium-scoped@luci-project-accounts.iam.gserviceaccount.com>	Tue Dec 14 09:54:52 2021
+tree	20c9ca1d8d2caf14be6eccf113550b929e2530e5
+parent	d5684ab99411bd973c4c5fcb400070c4094dd57d [diff]
+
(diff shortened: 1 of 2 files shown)


meta-chromium/recipes-browser/chromium/files/chromium-wayland/0001-ozone-wayland-fixed-terminate-caused-by-binding-to-wrong-version.patch
```diff
@@ -0,0 +1,568 @@
+author	Alexander Dunaev <adunaev@igalia.com>	Tue Dec 14 09:54:52 2021
```
msisov, line 1: Can you please add a proper description?

The template is the following -

Upstream-Status: Backport/Pending/Rejected

Short description if necessary

Signed-off-by: Name LastName <email@email.com>
---

See https://github.com/OSSystems/meta-browser/blob/master/meta-chromium/recipes-browser/chromium/files/chromium-wayland/0001-ozone-wayland-fix-re-initialization-of-WBMG.patch#L1-L8 for the concrete example.

MarkusVolk: Sure. Hope it's ok like this

msisov (approved):

meta-chromium/recipes-browser/chromium/files/chromium-wayland/0001-ozone-wayland-fixed-terminate-caused-by-binding-to-wrong-version.patch
msisov: yes. thanks.

rakuco:
Looks like this is ready to be merged. Thanks, @MarkusVolk

PR #602, opened 2022-01-24
chromium: update to 97.0.4692.71

msisov:
Release notes -
https://chromereleases.googleblog.com/2022/01/stable-channel-update-for-desktop.html

License changes:
---------------
No changes.
---------------

Test-built with:
* chromium-ozone-wayland:
  * honister,  clang, MACHINE="intel-corei7-64"
  * honister,  clang, MACHINE="raspberrypi3
  * hardknott, clang, MACHINE="raspberrypi3
  * dunfell, clang**, MACHINE="qemuarm64"
  * dunfell, clang**, MACHINE="m3ulcb"
    
* chromium-x11
  * honister,  clang, MACHINE="raspberrypi3"
  * hardknott, clang, MACHINE="raspberrypi3"
  * dunfell, clang**, MACHINE="raspberrypi3"
    
** Please note that Chromium requires clang version to be >= 12.
For that, when on dunfell branch, use meta-clang/clang12 branch.


msisov:
I haven't tested with dunfell as meta-clang/dunfell has clang version 10, which is pretty old. We need to update that. Otherwise, it becomes too time consuming to fix all these problems. The honister and hardknott use versions 13 and 12. In the mean time, Chromium uses clang 14. 

@kraj is it possible to update dunfell to at least clang version 12? I saw there was a branch called dunfell-clang12, but it doesn't seem to contain clang 12, but rather version 10.

rakuco:
I'm a bit surprised by the lack of license changes; could you just confirm that? The change itself looks fine to me otherwise.

msisov:
> I'm a bit surprised by the lack of license changes; could you just confirm that? The change itself looks fine to me otherwise.

I was also surprised, but the script that you shared with me long time ago produced a file with all the licenses, which diff didn't actually have any changes when compared with the licenses in the chromium.inc

msisov:
log
[log.txt](https://github.com/OSSystems/meta-browser/files/7925248/log.txt)

rakuco (approved):
lgtm. I'd like to hear from @kraj about dunfell and clang though, as landing this change as-is would break that branch that has worked so far.

rwmacleod:
Builds fine for the master branch for qemux86-64, qemuarm64.

msisov:
> Builds fine for the master branch for qemux86-64, qemuarm64.

thanks for testing!

@kraj gentle reminder

rakuco:
It looks like @kraj is currently away. cc'ing @otavio 

One option we have for now is to create a `dunfell` branch here that remains at the current HEAD (d25d8ee98a656b2534d0eec6138ef264529fab4f) while master progresses and is no longer guaranteed to work with dunfell. It unblocks this and future updates, but I'm guessing this will surprise and piss off our users on dunfell, and I'm not sure what to do if/when meta-clang's dunfell branch moves to a more recent clang release.

otavio:
If we could get meta-clang it would be great as it will avoid the other branch. @kraj would be the best one to comment. If it cannot be upgraded then the branch is the only option.

kraj:
I think its fine to use `dunfell_clang12` branch for dunfell

msisov:
> I think its fine to use `dunfell_clang12` branch for dunfell

Thanks. I just saw you pushed an update to that branch. I will try and report back

msisov:
@kraj it still uses clang10. Seems like the clang.inc needs to be updated as it has -

```
MAJOR_VER = "10"
MINOR_VER = "0"
PATCH_VER = "1"
```

kraj:
@msisov please cherry-pick https://github.com/YoeDistro/meta-browser/commit/931eedd84453994e39a22d747125b2953d5b91ed along this pull as well, this will make sure it works with musl as well.

rakuco:
> I think its fine to use `dunfell_clang12` branch for dunfell

Do you plan on merging dunfell_clang12 into dunfell in meta-clang? Requiring a different branch than what users are already using will likely cause confusion.

rakuco:
> > Do you plan on merging dunfell_clang12 into dunfell in meta-clang? Requiring a different branch than what users are already using will likely cause confusion.
> 
> Generally upgrading major compiler versions is not a good idea across a release. Even if we wanted to do so it should be added as an option with clang 10 being in there as default. then someone needs to test it for world builds with dunfell

Hmm, in this case what should we recommend to our users? Have separate meta-clang checkouts for chromium and the rest at different branches (not sure if this is possible)? Tell them to switch to the clang12 branch at their own risk?

msisov:
@kraj switching to the clang12 branch results int

```
ERROR: Nothing PROVIDES 'clang-native-x86_64' (but virtual:native:/home/msisov/code/yocto-rpi-dunfell/sources/meta-clang/recipes-devtools/clang/compiler-rt_git.bb DEPENDS on or otherwise requires it). Close matches:
  clang-native
  slang-native
ERROR: Required build target 'chromium-x11' has no buildable providers.
Missing or unbuildable dependency chain was: ['chromium-x11', 'gn-native', 'compiler-rt-native', 'clang-native-x86_64']
```

kraj:
> @kraj switching to the clang12 branch results int
> 
> ```
> ERROR: Nothing PROVIDES 'clang-native-x86_64' (but virtual:native:/home/msisov/code/yocto-rpi-dunfell/sources/meta-clang/recipes-devtools/clang/compiler-rt_git.bb DEPENDS on or otherwise requires it). Close matches:
>   clang-native
>   slang-native
> ERROR: Required build target 'chromium-x11' has no buildable providers.
> Missing or unbuildable dependency chain was: ['chromium-x11', 'gn-native', 'compiler-rt-native', 'clang-native-x86_64']
> ```
@msisov 
yes it needs some fixed from `dunfell` can you try with `kraj/clang12` branch where I have added the needed merges.

barndoor2276:
> > @kraj switching to the clang12 branch results int
> > ```
> > ERROR: Nothing PROVIDES 'clang-native-x86_64' (but virtual:native:/home/msisov/code/yocto-rpi-dunfell/sources/meta-clang/recipes-devtools/clang/compiler-rt_git.bb DEPENDS on or otherwise requires it). Close matches:
> >   clang-native
> >   slang-native
> > ERROR: Required build target 'chromium-x11' has no buildable providers.
> > Missing or unbuildable dependency chain was: ['chromium-x11', 'gn-native', 'compiler-rt-native', 'clang-native-x86_64']
> > ```
> 
> @msisov yes it needs some fixed from `dunfell` can you try with `kraj/clang12` branch where I have added the needed merges.

I tried a build with `kraj/clang12` branch and ended up with an ungodly long compilation error in `clang-native-12.0.0 do_compile`. Log file attached.
[log.txt](https://github.com/OSSystems/meta-browser/files/8024496/log.txt)

msisov:
> > @kraj switching to the clang12 branch results int
> > ```
> > ERROR: Nothing PROVIDES 'clang-native-x86_64' (but virtual:native:/home/msisov/code/yocto-rpi-dunfell/sources/meta-clang/recipes-devtools/clang/compiler-rt_git.bb DEPENDS on or otherwise requires it). Close matches:
> >   clang-native
> >   slang-native
> > ERROR: Required build target 'chromium-x11' has no buildable providers.
> > Missing or unbuildable dependency chain was: ['chromium-x11', 'gn-native', 'compiler-rt-native', 'clang-native-x86_64']
> > ```
> 
> @msisov yes it needs some fixed from `dunfell` can you try with `kraj/clang12` branch where I have added the needed merges.

same problem -

```
ERROR: Nothing PROVIDES 'clang-native-x86_64' (but virtual:native:/home/msisov/code/yocto-rpi-dunfell/sources/meta-clang/recipes-devtools/clang/compiler-rt_git.bb DEPENDS on or otherwise requires it). Close matches:
  clang-native
  slang-native
ERROR: Required build target 'chromium-x11' has no buildable providers.
Missing or unbuildable dependency chain was: ['chromium-x11', 'gn-native', 'compiler-rt-native', 'clang-native-x86_64']
```

kraj:
@msisov perhaps we need to switch to `dunfell_clang12` but this branch is not uptodate with `dunfell`, so I am trying to merge them via `kraj/clang12` but guess its lacking a bit, let me spend some time looking into it.

kraj:
@msisov I have pushed changes for clang12 on dunfell to `kraj/clang12` branch. I am able to build `chromium-ozone-wayland-97.0.4692.71-r0` this pull request on dunfell + kraj/clang12 combo, please give it a shot, if it works, I will merge it into `dunfell_clang12`

msisov:
@kraj thanks. the build succeeded. Let me know when I can merge this chromium update.

rakuco:
> > > Do you plan on merging dunfell_clang12 into dunfell in meta-clang? Requiring a different branch than what users are already using will likely cause confusion.
> > 
> > 
> > Generally upgrading major compiler versions is not a good idea across a release. Even if we wanted to do so it should be added as an option with clang 10 being in there as default. then someone needs to test it for world builds with dunfell
> 
> Hmm, in this case what should we recommend to our users? Have separate meta-clang checkouts for chromium and the rest at different branches (not sure if this is possible)? Tell them to switch to the clang12 branch at their own risk?

I'd still appreciate an answer to the above.

kraj:
> > > > Do you plan on merging dunfell_clang12 into dunfell in meta-clang? Requiring a different branch than what users are already using will likely cause confusion.
> > > 
> > > 
> > > Generally upgrading major compiler versions is not a good idea across a release. Even if we wanted to do so it should be added as an option with clang 10 being in there as default. then someone needs to test it for world builds with dunfell
> > 
> > 
> > Hmm, in this case what should we recommend to our users? Have separate meta-clang checkouts for chromium and the rest at different branches (not sure if this is possible)? Tell them to switch to the clang12 branch at their own risk?
> 
> I'd still appreciate an answer to the above.

Clang12 branch is a bit better tested with dunfell ecosystem since it was submitted by Huawei devs 
So it might be better than a general major upgrade

kraj:
> @kraj thanks. the build succeeded. Let me know when I can merge this chromium update.

its ready to merge. We still need https://github.com/YoeDistro/meta-browser/commit/931eedd84453994e39a22d747125b2953d5b91ed

but I can post that as followup.

msisov:
> > > > > Do you plan on merging dunfell_clang12 into dunfell in meta-clang? Requiring a different branch than what users are already using will likely cause confusion.
> > > > 
> > > > 
> > > > Generally upgrading major compiler versions is not a good idea across a release. Even if we wanted to do so it should be added as an option with clang 10 being in there as default. then someone needs to test it for world builds with dunfell
> > > 
> > > 
> > > Hmm, in this case what should we recommend to our users? Have separate meta-clang checkouts for chromium and the rest at different branches (not sure if this is possible)? Tell them to switch to the clang12 branch at their own risk?
> > 
> > 
> > I'd still appreciate an answer to the above.
> 
> Clang12 branch is a bit better tested with dunfell ecosystem since it was submitted by Huawei devs So it might be better than a general major upgrade

Probably, it'll make sense to switch meta-clang/dunfell to clang12 to avoid confusions?

rakuco:
> Probably, it'll make sense to switch meta-clang/dunfell to clang12 to avoid confusions before merging this?

From a meta-browser perspective, absolutely, otherwise the problem of having to tell users to switch to a different meta-clang branch will confuse a lot of people. @kraj had originally said switching to a new compiler version in a well-established branch like dunfell would be dangerous, but the latest comment says it might be possible to do that?

kraj:
Use dunfell_clang12 branch for now the if this will be merged into main dunfell will take some time and consideration

msisov:
@rakuco wdyt? I have mixed feelings about that. Of course, I can add that to the commit message. But I suspect most users will ignore that.

rakuco:
> @rakuco wdyt? I have mixed feelings about that. Of course, I can add that to the commit message. But I suspect most users will ignore that.

Yeah, it's definitely not ideal. It's been a while since I had to write custom tasks in Bitbake, but I wonder if there's a way to check clang's version and fail with an explicit error message telling users to switch their branch if it's e.g. < 12.0.

msisov:
> > @rakuco wdyt? I have mixed feelings about that. Of course, I can add that to the commit message. But I suspect most users will ignore that.
> 
> Yeah, it's definitely not ideal. It's been a while since I had to write custom tasks in Bitbake, but I wonder if there's a way to check clang's version and fail with an explicit error message telling users to switch their branch if it's e.g. < 12.0.

I only know about PREFERRED_VERSION, but I don't think it works the way we'd want it to work. Is there an equivalent of that, but for a required version?

rakuco:
meta-clang's conf/layer.conf has `LLVMVERSION`, maybe something like this works?

```
python do_check_llvm_version () {
  llvm_major_version = int((d.getVar("LLVMVERSION") or "").split()[0])  # probably with better error checking
  if llvm_major_version < 12:
    bb.fatal('Uh-oh, your LLVM Is too old, if you are on dunfell make sure you use the clang12 branch')
}
addtask check_llvm_version before do_configure
```

kraj:
> meta-clang's conf/layer.conf has `LLVMVERSION`, maybe something like this works?
> 
> ```
> python do_check_llvm_version () {
>   llvm_major_version = int((d.getVar("LLVMVERSION") or "").split()[0])  # probably with better error checking
>   if llvm_major_version < 12:
>     bb.fatal('Uh-oh, your LLVM Is too old, if you are on dunfell make sure you use the clang12 branch')
> }
> addtask check_llvm_version before do_configure
> ```
yeah you can process it from LLVMVERSION

msisov:
> > meta-clang's conf/layer.conf has `LLVMVERSION`, maybe something like this works?
> > ```
> > python do_check_llvm_version () {
> >   llvm_major_version = int((d.getVar("LLVMVERSION") or "").split()[0])  # probably with better error checking
> >   if llvm_major_version < 12:
> >     bb.fatal('Uh-oh, your LLVM Is too old, if you are on dunfell make sure you use the clang12 branch')
> > }
> > addtask check_llvm_version before do_configure
> > ```
> 
> yeah you can process it from LLVMVERSION

good idea. done.

If it is unable to determine the LLVMVERSION, the following warning is printed -
"WARNING: chromium-ozone-wayland-97.0.4692.71-r0 do_check_llvm_version: Unable to determine LLVM version. Chromium may not be compiled successfully if the LLVM version is below 12.0.0."

If the version is below 12.0.0, the following fatal error is printed - 
"ERROR: chromium-ozone-wayland-97.0.4692.71-r0 do_check_llvm_version: Your LLVMVERSION (10.0.1) is lower than the minimum required LLVMVERSION (12.0.0). If you are using dunfell, make sure you use clang12 branch.

rakuco:
@msisov still lgtm, thank you and @kraj for working on this.

Can you just do two things before merging:
* Add your dunfell tests to the list in the commit/PR message
* Mention that dunfell users need the clang12 branch in the commit/PR message

msisov:
> @msisov still lgtm, thank you and @kraj for working on this.
> 
> Can you just do two things before merging:
> 
> * Add your dunfell tests to the list in the commit/PR message
> * Mention that dunfell users need the clang12 branch in the commit/PR message

thanks for noting that. done.

kraj (approved):

PR #603, opened 2022-02-04
chromium: don't pass is_official_build=true when doing a component-build

Villemoes:
When setting the PACKAGECONFIG component-build, the build breaks
immediately due to this from the toplevel BUILD.gn:

if (is_official_build) {
  # An official (maximally optimized!) component (optimized for build times)
  # build doesn't make sense and usually doesn't work.
  assert(!is_component_build)
}

So we must make is_official_build the negative of component-build.

### DIFF
--- a/meta-chromium/recipes-browser/chromium/chromium-gn.inc
+++ b/meta-chromium/recipes-browser/chromium/chromium-gn.inc
@@ -165,7 +165,8 @@ GN_ARGS += 'host_pkg_config="pkg-config-native"'
 # (debug, release, official) but for historical reasons there are two
 # separate flags.
 # See also: https://groups.google.com/a/chromium.org/d/msg/chromium-dev/hkcb6AOX5gE/PPT1ukWoBwAJ
-GN_ARGS += "is_debug=false is_official_build=true"
+GN_ARGS += "is_debug=false"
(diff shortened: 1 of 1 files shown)


Villemoes:
Ping. Is there any other way to get the component-build packageconfig working? It is officially listed in the README.md, but does it actually currently work for anybody?

rakuco:
The patch looks valid to me, but it does seem to indicate that nobody has tested the component build knob in ages. Do you get a working binary if you build the recipe with this patch?

Villemoes:
Unfortunately, no, I don't even get a binary. The build fails towards the end with an error I can't decipher:
```
./v8_context_snapshot_generator: symbol lookup error: ...out/Release/libbase_i18n.so: undefined symbol: UCNV_FROM_U_CALLBACK_SKIP_71
./v8-qemu-wrapper.sh failed with exit code 127
```
But that symbol does exist in `libicuuc.so`, and `libbase_i18n.so` itself has been succesfully created (and libicuuc.so is in the solibs= for libbase_i18n.so, and also listed as NEEDED when I run `readelf -d` on it). So I suppose this is some defect in the v8_context_snapshot_generator tool, but I have no idea how to proceed or fix it.

PR #615, opened 2022-03-06
chromium,firefox: Use SPDX naming scheme for LICENSE

kraj:

### DIFF
--- a/meta-chromium/recipes-browser/chromium/chromium.inc
+++ b/meta-chromium/recipes-browser/chromium/chromium.inc
@@ -53,7 +53,7 @@ DEPENDS += "python3-setuptools-native"
 # LICENSE does not cover all the licenses in LIC_FILES_CHKSUM; instead, it
 # currently contains Chromium's license (BSD-3-Clause) as well as Blink's
 # (LGPL2.0, LGPL2.1 and BSD-2-Clause).
-LICENSE = "BSD-3-Clause & LGPL-2.0-only & LGPL-2.1-only"
+LICENSE = "BSD-3-Clause & LGPL-2.0-or-later & LGPL-2.1-or-later"
 LIC_FILES_CHKSUM = "\
     file://${S}/LICENSE;md5=0fca02217a5d49a14dfe2d11837bb34d \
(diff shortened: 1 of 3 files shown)


meta-chromium/recipes-browser/chromium/chromium.inc
```diff
@@ -53,7 +53,7 @@ DEPENDS += "python3-setuptools-native"
 # LICENSE does not cover all the licenses in LIC_FILES_CHKSUM; instead, it
 # currently contains Chromium's license (BSD-3-Clause) as well as Blink's
 # (LGPL2.0, LGPL2.1 and BSD-2-Clause).
-LICENSE = "BSD-3-Clause & LGPL-2.0 & LGPL-2.1"
+LICENSE = "BSD-3-Clause & LGPL-2.0-only & LGPL-2.1-only"
```
rakuco, line 56: I think the `-or-later` version of the LGPL parts makes more sense. I see several occurrences of _"version 2, or (at your option) any later version"_ (and the 2.1 equivalent) in Blink headers, for example.

meta-chromium/recipes-browser/chromium/chromium.inc
```diff
@@ -53,7 +53,7 @@ DEPENDS += "python3-setuptools-native"
 # LICENSE does not cover all the licenses in LIC_FILES_CHKSUM; instead, it
 # currently contains Chromium's license (BSD-3-Clause) as well as Blink's
 # (LGPL2.0, LGPL2.1 and BSD-2-Clause).
-LICENSE = "BSD-3-Clause & LGPL-2.0 & LGPL-2.1"
+LICENSE = "BSD-3-Clause & LGPL-2.0-only & LGPL-2.1-or-later"
```
rakuco, line 56: The LGPL-2.0 part should be "or-later" too, I think. Like I said in the other review comment:

> I see several occurrences of "version 2, or (at your option) any later version" (and the 2.1 equivalent) in Blink headers, for example.

rakuco (approved):
thanks!

PR #619, opened 2022-03-12
firefox: fix build with kirkstone

shr-project:
I haven't tested this in runtime, the only motivation for this PR was to measure build time with kirkstone for https://github.com/shr-project/test-oe-build-time, nobody should be using this old firefox on embedded devices.

### DIFF
--- a/meta-firefox/classes/rust-common-without-python3native.bbclass
+++ b/meta-firefox/classes/rust-common-without-python3native.bbclass
@@ -1,0 +1,186 @@
+# removed, because firefox fails to configure when it finds native python3 instead of python2.7 from HOSTTOOLS
+# inherit python3native
+
+# Common variables used by all Rust builds
+export rustlibdir = "${libdir}/rust"
+FILES:${PN} += "${rustlibdir}/*.so"
+FILES:${PN}-dev += "${rustlibdir}/*.rlib ${rustlibdir}/*.rmeta"
+FILES:${PN}-dbg += "${rustlibdir}/.debug"
+
+RUSTLIB = "-L ${STAGING_LIBDIR}/rust"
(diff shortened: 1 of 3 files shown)


meta-firefox/classes/rust-common-without-python3native.bbclass
```diff
@@ -0,0 +1,186 @@
...
+	create_wrapper "${RUST_TARGET_CXX}" "${WRAPPER_TARGET_CXX}"
+	# Yocto Target / Rust Target linker
+	create_wrapper "${RUST_TARGET_CCLD}" "${WRAPPER_TARGET_CCLD}" "${WRAPPER_TARGET_LDFLAGS}"
+	# Yocto Target / Rust Target archiver
+	create_wrapper "${RUST_TARGET_AR}" "${WRAPPER_TARGET_AR}"
+
+}
+
+addtask rust_create_wrappers before do_configure after do_patch do_prepare_recipe_sysroot
+do_rust_create_wrappers[dirs] += "${WRAPPER_DIR}"
```
kraj, line 186: I think its worth bringing this to notice of oe-core and see if this undoing is not needed, firefox is one of main consumers of rust and I wonder if the oe-core change is done to support some oe-core app usecase which breaks firefox and perhaps a lot more of other rust using apps.

shr-project: in this case I would say that it's completely firefox fault, that detecting python3 from PYTHON variable breaks its ancient configure script, firefox isn't well maintained (I did all the recent updates only to measure build time) and inherit python3native should be reasonably safe anywhere else (through rust-common.bbclass or any other way). python + python2.7 in HOSTTOOLS from meta-firefox's layer.conf is much worse evil than this.

kraj: I see thanks

kraj (approved):
lgtm

PR #621, opened 2022-03-18
chromium: Ignore textrels on x86 build

kraj:
building 32bit x86 results in textrels which I guess should be fixed
so here is a workaround until they are root caused and fixed.

### DIFF
--- a/meta-chromium/recipes-browser/chromium/chromium-gn.inc
+++ b/meta-chromium/recipes-browser/chromium/chromium-gn.inc
@@ -521,3 +521,7 @@ PACKAGE_DEBUG_SPLIT_STYLE = "debug-without-src"
 
 # There is no need to ship empty -dev packages.
 ALLOW_EMPTY:${PN}-dev = "0"
+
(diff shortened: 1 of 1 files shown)


meta-chromium/recipes-browser/chromium/chromium-gn.inc
```diff
@@ -521,3 +521,5 @@ PACKAGE_DEBUG_SPLIT_STYLE = "debug-without-src"
 
 # There is no need to ship empty -dev packages.
 ALLOW_EMPTY:${PN}-dev = "0"
+INSANE_SKIP:${PN}:append:x86 = "textrel"
+
```
rakuco, line 525: Can you add a comment above the code explaining what this is working around? Like a sample error message.

kraj: done

rakuco: Thanks, I see you added it to the commit message, but I meant a comment right above this line. Can you add it as well?

kraj: done

rakuco (approved):
lgtm!

Issue #624, opened 2022-04-02
chrome_crashpad_handler: symbol lookup error: /lib/libc++abi.so.1: undefined symbol: __cxa_end_cleanup


MastaG:
I'm building chromium-ozone-wayland using the honister branches of openembedded-core and meta-openembedded and meta-clang with bitbake 1.52.

But after upgrading my OS to Fedora 36 (armv7) it doesn't seem to run any longer.
The crashpad_handler seems to fail with:
```
/usr/local/lib/chromium-wayland/chrome_crashpad_handler: symbol lookup error: /lib/libc++abi.so.1: undefined symbol: __cxa_end_cleanup
```
Now it seems honister builds using LLVM/clang 13, but Fedora 36 seems to ship mixed versions:
```
libcxxabi-14.0.0-1.fc36.armv7hl
libcxx-14.0.0-1.fc36.armv7hl
llvm-libs-14.0.0-1.fc36.armv7hl
llvm13-libs-13.0.1-1.fc36.armv7hl
```
It does provide version 13 for the LLVM libraries, but libcxx is only available as version 14 now.
But it seems the library does contain the symbol:
```
$ grep __cxa_end_cleanup /lib/libc++abi.so.1
grep: /lib/libc++abi.so.1: binary file matches
$ rpm -qf /lib/libc++abi.so.1
libcxxabi-14.0.0-1.fc36.armv7hl
```

Any ideas on how to solve this?

kraj:
are you building chromium using OE/Yocto and then running it with fedora runtime ? If so, I am afraid this is not much tested and even supported combo. You should also use runtime from the oe/yocto builds. We have no control over how fedora would have compiled with llvm runtime.

MastaG:
Yes that's correct. I only use oe/yocto to build it and then move it to my Fedora installation.
Would it be possible to link libc++ and libc++abi statically instead?
I believe the original Chrome browser does this as well?

kraj:
> Yes that's correct. I only use oe/yocto to build it and then move it to my Fedora installation. Would it be possible to link libc++ and libc++abi statically instead? I believe the original Chrome browser does this as well?

that is possible and perhaps what we could do as long as we don't call C++ system libraries we should be fine, although this will mean a bit of duplication of libraries in case folks are already using LLVM runtime with yocto

anyway, I have posted a patch to make this setting as a packageconfig. So you can try it out to enable this knob in settings and report back

https://github.com/OSSystems/meta-browser/pull/625

[...]

PR #625, opened 2022-04-03
chromium: Create packageconfig to control libc++ to use

kraj:
This helps to use either system provided libc++ or bundled libc++ with
chromium. Defaults in chromium is to use bundled version, however in OE
we have used the system provided so far, therefore the default are not
changed here, just an option offered to enable custom libc++ if someone
desires so. Should we default to use the custom copy by default, perhaps
yes but thats a separate discussion

### DIFF
--- a/meta-chromium/README.md
+++ b/meta-chromium/README.md
@@ -90,6 +90,12 @@ PACKAGECONFIG knobs
 * cups: (off by default)
   Enables CUPS support in Chromium, and adds a dependency on the "cups" recipe.
 
+* custom-libcxx (off by default)
+  Enable vendored version of C++ runtime ( libc++ ) instead of using this from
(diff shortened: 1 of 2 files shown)


MastaG:
As I pointed out in my issue: https://github.com/OSSystems/meta-browser/issues/624
The packageconfig flag to build using the custom libcxx works great.
It only seems to depend on llvm's libunwind.so.1 but that's only a problem on my side (as I move the chromium installation to Fedora OS).
So I guess this can be merged.

kraj:
@rakuco can you review this change for me ?

rakuco (approved):
lgtm provided you document the new knob.

rakuco (approved):
lgtm, thanks

PR #629, opened 2022-04-15
chromium: Fix build with musl

kraj:
chromium version upgrades means the musl patches also needs to be
forward ported or removed if not needed.

Tested on rpi4-64


rakuco (approved):
lgtm

PR #673, opened 2022-11-06
tools/grit/grit/util.py: remove deprecated mode for open

MarkusVolk:
Signed-off-by: Markus Volk <f_l_k@t-online.de>


rakuco (commented):
Thank you for the patch. This has already been fixed upstream, so could you update your patch metadata and mention it's a backport and point to the upstream commit (https://chromium.googlesource.com/chromium/src/+/0991fc6acd3c85472000f2055af542515c3c6297%5E%21/ or https://chromium-review.googlesource.com/c/chromium/src/+/3859535)?

MarkusVolk:
Thanks for the review. I had not yet noticed that the problem has already been fixed

rakuco (approved):
lgtm, thank you!

PR #678, opened 2022-11-16
chromium-x11: fix build error due to missing wayland-client

twoerner:
The following error occurs during do_configure and is fixed by adding a dependency on wayland... for the chromium-x11 build:

	| ERROR at //build/config/linux/pkg_config.gni:104:17: Script returned non-zero exit code.
	|     pkgresult = exec_script(pkg_config_script, args, "value")
	|                 ^----------
	| Current dir: /z/build-master/6951/seccamviewer/build/tmp-glibc/work/cortexa53-oe-linux/chromium-x11/106.0.5249.119-r0/chromium-106.0.5249.119/out/Release/
	| Command: python3 /z/build-master/6951/seccamviewer/build/tmp-glibc/work/cortexa53-oe-linux/chromium-x11/106.0.5249.119-r0/chromium-106.0.5249.119/build/config/linux/pkg-config.py wayland-client
	| Returned 1.
	| stderr:
	|
	| Package wayland-client was not found in the pkg-config search path.
	| Perhaps you should add the directory containing `wayland-client.pc'
	| to the PKG_CONFIG_PATH environment variable
	| No package 'wayland-client' found
	| Could not run pkg-config.
	|
	| See //third_party/wayland/BUILD.gn:172:3: whence it was called.
	|   pkg_config("wayland_client_config") {
	|   ^------------------------------------
	| See //third_party/swiftshader/src/WSI/BUILD.gn:92:15: which caused the file to be included.
	|     deps += [ "$wayland_gn_dir:wayland_client" ]

Build and run tested on raspberrypi4-64.

Signed-off-by: Trevor Woerner <twoerner@gmail.com>


MastaG:
So swiftshader now seems to depend on wayland?

MastaG:
https://github.com/JeffyCN/meta-rockchip/blob/master/dynamic-layers/recipes-browser/chromium/chromium_105.0.5195/0018-swiftshader-Don-t-check-system-wayland-for-non-wayla.patch

rakuco:
Thanks, @MastaG. It looks like this was also fixed upstream by https://swiftshader-review.googlesource.com/c/SwiftShader/+/68489 which does the same thing as the linked patch

@twoerner can you update the PR and backport that change instead?

twoerner:
Yes, I'll go ahead and do that.

Thanks @MastaG I suspected there was a better solution

twoerner:
Updated and ready for re-evaluation.

meta-chromium/recipes-browser/chromium/chromium-gn.inc
```diff
@@ -25,6 +25,10 @@ SRC_URI += " \
     file://0014-grit-util-py-remove-deprecated-mode-for-open.patch \
 "
 
+SRC_URI:append:pn-chromium-x11 = " \
```
rakuco, line 28: You can drop the `pn` part here and just add the patch unconditionally. chromium-ozone-wayland sets `ozone_platform_wayland=true` so the patch doesn't break anything there.

twoerner:
Updated. Thanks for the review @rakuco

meta-chromium/recipes-browser/chromium/files/0018-swiftshader-Don-t-check-system-wayland-for-non-wayla.patch
```diff
@@ -0,0 +1,27 @@
+From 8e61ec315214833b71f75e463919626fd519af48 Mon Sep 17 00:00:00 2001
+From: Jeffy Chen <jeffy.chen@rock-chips.com>
```
rakuco, line 2: One last thing: instead of importing the patch from meta-rockchip, can you use the upstream commit message and add an "Upstream-status" tag? The upstream change also adds the `ozone_platform_wayland` check to the middle below rather than at the end.

Doing so makes it easier for us to remove this change once we move to a newer release.

twoerner: Okay, putting that together now.

twoerner: ...and now for the 4h build...

rakuco (approved):
lgtm, thank you.

I've added two new commits on top of yours with some small adjustments:
* Add the missing "Upstream-Status" tag to the patch
* Rename the patch from 0001-foo to 0015-foo

rakuco:
Oops, my plan was to squash everything into a single one, but the repository only allows me to rebase and merge. @twoerner please let me know if you prefer to squash everything and force-push or if I should just go ahead.

twoerner:
Go ahead and apply. I won't have time to look at this today.

PR #690, opened 2023-01-26
chromium-x11: Add RDEPENDS on libx11-xcb.

### LINKED ISSUE #683: chromium-x11 RDEPENDS is missing libx11-xcb
chromium-x11 dependency list should also include:

`libx11-xcb`

otherwise if the `libX11-xcb.so.1` is not present on the target, chromium won't start.

https://github.com/OSSystems/meta-browser/blob/master/meta-chromium/recipes-browser/chromium/chromium-x11_106.0.5249.119.bb#L5

A workaround is to explicitly add `libx11-xcb` to `IMAGE_INSTALL` in `local.conf`.

rakuco:
libX11-xcb.so.1 is loaded at runtime by the X11 backend, but we had no
explicit dependency on the package that provides it (libx11-xcb).

It ended up being installed most of the time anyway, but make sure it is
listed as a runtime dependency to avoid the issue described in #683.

Fixes #683.

### DIFF
--- a/meta-chromium/recipes-browser/chromium/chromium-x11_109.0.5414.74.bb
+++ b/meta-chromium/recipes-browser/chromium/chromium-x11_109.0.5414.74.bb
@@ -16,6 +16,10 @@ DEPENDS += "\
         libxtst \
 "
 
+# Loaded at runtime.
+# https://source.chromium.org/chromium/chromium/src/+/main:ui/gfx/x/xlib_support.cc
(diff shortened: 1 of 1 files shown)


rakuco:
cc @MastaG @royaltm
I haven't tested it myself, but @MastaG's suggestion makes sense given how the library is loaded at runtime.

kraj (approved):
lgtm

MastaG:
Lgtm2

PR #692, opened 2023-02-09
chromium: update generate_license_list.py

MaxIhlenfeldt:
Chromium's `license.py` moved from `tools/` to `tools/licenses/` [1]. Our own `generate_license_list.py` needs to use this new path.

[1] https://chromium-review.googlesource.com/c/chromium/src/+/4022205

### DIFF
--- a/meta-chromium/recipes-browser/chromium/generate_license_list.py
+++ b/meta-chromium/recipes-browser/chromium/generate_license_list.py
@@ -1,8 +1,7 @@
 #!/usr/bin/env python3
 """
 This script can be used to generate LIC_FILES_CHKSUM in chromium.inc.
-
-It uses Chromium's own tools/licenses.py script to scan for third_party
+It uses Chromium's own tools/licenses/licenses.py script to scan for third_party
 directories and license files. This means its output is generated on a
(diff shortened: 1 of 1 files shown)


rakuco:
The upstream change, https://chromium-review.googlesource.com/c/chromium/src/+/4022205, seems to have happened in M110.

Is there a particular reason not to have this patch as part of the recipe update? If we merge this PR as-is, the script will be broken for the version we currently ship.

meta-chromium/recipes-browser/chromium/generate_license_list.py
```diff
@@ -1,8 +1,7 @@
 #!/usr/bin/env python3
 """
 This script can be used to generate LIC_FILES_CHKSUM in chromium.inc.
-
```
rakuco, line 4: Can you add this empty line back? The format here was following https://peps.python.org/pep-0257/#multi-line-docstrings.

MaxIhlenfeldt:
> Is there a particular reason not to have this patch as part of the recipe update? If we merge this PR as-is, the script will be broken for the version we currently ship.

You're right, that's better. I'll include it in the recipe update :+1:

PR #695, opened 2023-02-21
chromium: update to 110.0.5481.77

MaxIhlenfeldt:
Release notes:
    https://chromereleases.googleblog.com/2023/02/stable-channel-update-for-desktop.html

Build and patch changes:
------------------------

Add DEPENDS on libffi as Chromium doesn't allow building with system libwayland anymore, and needs libffi to build its own libwayland.

Add libffi_%.bbappend to also copy the PIC version of libffi.a to the sysroot, as Chromium needs it. This is the same approach that Debian takes, which Chromium uses for its sysroot:
    https://sources.debian.org/src/libffi/3.4.4-1/debian/rules/#L103

Remove angle_use_wayland build arg as Chromium now uses Wayland >= 1.20.

Update generate_license_list.py to adapt to:
    2be00b963986 Move license tooling into //tools/licenses

Merge 0010-Revert-ios-Fix-breakpoint-in-Swift-files.patch with 0011-Revert-Use-ffile-compilation-dir-instead-of-fdebug-c.patch, as proposed in the former patch's commit message:
  > This should be fixed by merging these two patches in the next rebase
  > and the changes to toolchain/ios/swiftc.py dropped.

Drop backported 0012-Build-X11-parts-only-for-ozone_platform_x11.patch, which was already removed from the build in the update to 109.

Update 0013-Fix-html_minifier-script-for-node-v12.patch's Upstream-Status to Inappropriate, per discussion with the maintainer:
    https://crrev.com/c/4239697.

Drop backported 0014-blink-Fix-includes-for-metronome_source.patch.

Rebase remaining patches.

License changes:
----------------

Added licenses:
  * ios/third_party/lottie/LICENSE
  - https://chromium-review.googlesource.com/c/chromium/src/+/4024225

  * third_party/android_deps/libs/android_arch_core_common/LICENSE
  - https://chromium-review.googlesource.com/837189 (added), https://chromium-review.googlesource.com/1142189 (renamed)

  * third_party/android_deps/libs/com_google_android_annotations/LICENSE
  * third_party/android_deps/libs/io_grpc_grpc_api/LICENSE
  * third_party/android_deps/libs/io_grpc_grpc_binder/LICENSE
  * third_party/android_deps/libs/io_grpc_grpc_context/LICENSE
  * third_party/android_deps/libs/io_grpc_grpc_core/LICENSE
  * third_party/android_deps/libs/io_grpc_grpc_protobuf_lite/LICENSE
[...]

PR #697, opened 2023-02-28
chromium: incorporate upstream ANGLE and Swiftshader changes

MaxIhlenfeldt:
As discovered by @rakuco during the discussion off #695, we missed some upstream changes to what libraries are built and installed for ANGLE and Swiftshader:
 * https://chromium-review.googlesource.com/c/chromium/src/+/3380402 Removed legacy Swiftshader. We don't need to install libEGL.so and libGLESv2.so to chromium/swiftshader anymore.
 * https://chromium-review.googlesource.com/c/chromium/src/+/3015696 Install libvulkan.so.1, which ANGLE needs.
 * https://chromium-review.googlesource.com/c/chromium/src/+/2845811 Install Swiftshader VK files.

### DIFF
--- a/meta-chromium/recipes-browser/chromium/chromium-gn.inc
+++ b/meta-chromium/recipes-browser/chromium/chromium-gn.inc
@@ -475,12 +475,22 @@ do_install() {
 	# modifying the dummy "CHROME_EXTRA_ARGS" line
 	sed -i "s/^CHROME_EXTRA_ARGS=\"\"/CHROME_EXTRA_ARGS=\"${CHROMIUM_EXTRA_ARGS}\"/" ${D}${libdir}/chromium/chromium-wrapper
 
-	# This is ANGLE, not to be confused with the similarly named files under swiftshader/
+	# ANGLE.
 	if [ -e libEGL.so ]; then
-		install -m 0755 libEGL.so ${D}${libdir}/chromium/
+		install -m 0644 libEGL.so ${D}${libdir}/chromium/
+		install -m 0644 libGLESv2.so ${D}${libdir}/chromium/
 	fi
-	if [ -e libGLESv2.so ]; then
-		install -m 0755 libGLESv2.so ${D}${libdir}/chromium/
+
+	# libvulkan (for ANGLE).
+	if [ -e libvulkan.so.1 ]; then
+		install -m 0644 libvulkan.so.1 ${D}${libdir}/chromium/
+	fi
+
+	# Swiftshader VK.
+	if [ -e libvk_swiftshader.so ]; then
+		install -m 0644 libvk_swiftshader.so ${D}${libdir}/chromium/
+		# This is needed for ANGLE to find libvk_swiftshader.so.
+		install -m 0644 vk_swiftshader_icd.json ${D}${libdir}/chromium/
 	fi
 
 	if [ -n "${@bb.utils.contains('PACKAGECONFIG', 'component-build', 'component-build', '', d)}" ]; then
@@ -492,13 +502,6 @@ do_install() {
 	# So we simply check whether it exists in all cases and ship it.
 	if [ -e libminigbm.so ]; then
 		install -m 0755 libminigbm.so ${D}${libdir}/chromium/
-	fi
-
-	# Swiftshader is only built for x86 and x86-64.
-	if [ -d "swiftshader" ]; then
-		install -d ${D}${libdir}/chromium/swiftshader
-		install -m 0644 swiftshader/libEGL.so ${D}${libdir}/chromium/swiftshader/
-		install -m 0644 swiftshader/libGLESv2.so ${D}${libdir}/chromium/swiftshader/
 	fi
 
 	# ChromeDriver.


meta-chromium/recipes-browser/chromium/chromium-gn.inc
```diff
@@ -468,19 +468,33 @@ do_install() {
 	install -m 0644 chrome_*.pak ${D}${libdir}/chromium/
 	install -m 0644 resources.pak ${D}${libdir}/chromium/resources.pak
 
-	# Locales.
+	# Locales
```
rakuco, line 471: nit: this looks unrelated.

meta-chromium/recipes-browser/chromium/chromium-gn.inc
```diff
@@ -468,19 +468,33 @@ do_install() {
...
+	if [ -e libvulkan.so.1 ]; then
+		install -m 0755 libvulkan.so.1 ${D}${libdir}/chromium/
+	fi
+
+	# Swiftshader VK
+	if [ -e libvk_swiftshader.so ]; then
+		install -m 0644 libvk_swiftshader.so ${D}${libdir}/chromium/
+	fi
+	# This is needed for ANGLE to find libvk_swiftshader.so.
+	if [ -e vk_swiftshader_icd.json ]; then
```
rakuco, line 496: I wonder if we can do the same thing as https://source.chromium.org/chromium/chromium/src/+/main:chrome/installer/linux/common/installer.include;l=234;drc=defe5e193d6e2a35f49f098c19ac4a4b508652a5 and always assume vk_swiftshader_icd.json is present if libvk_swiftshader.so exists.

rakuco (approved):
lgtm with a few nits, but please wait for @rwmacleod before merging.

The permissions being used to install the files are a bit mixed up, but apparently upstream has the same issue :/

meta-chromium/recipes-browser/chromium/chromium-gn.inc
MaxIhlenfeldt: It's a habit of mine to clean up a bit if I find things in files I'm already working on. In this case, I think it makes the comment more consistent with other comments, but if you disagree (with this change in particular or the habit in general) I can leave the comment as-is.

MaxIhlenfeldt:
> please wait for @rwmacleod before merging.

Ack.

> The permissions being used to install the files are a bit mixed up, but apparently upstream has the same issue :/

Yes, I tried to use the same permission that the upstream install script uses. I hope I got them right...

meta-chromium/recipes-browser/chromium/chromium-gn.inc
MaxIhlenfeldt: Given that they are (currently, at least) [always built together](https://source.chromium.org/chromium/chromium/src/+/main:third_party/swiftshader/BUILD.gn;l=70-71;drc=166966576aeed4d4e6e822f707e710022a94f709), I suppose that assumption is reasonable, wdyt? I just thought from looking at the rest of the existing install script that the preferred way is to check each file separately.

meta-chromium/recipes-browser/chromium/chromium-gn.inc
rakuco: In general, I prefer not to touch unrelated parts of the code in a change to make it simpler for reviewers and complicating `git blame` calls. I usually try to follow Chromium's/Google's guideline of ending sentences with a period, but I guess over time several comments slipped in without one.

meta-chromium/recipes-browser/chromium/chromium-gn.inc
rakuco: I think it makes our lives easier if we follow what upstream does and always assume both files exist if the shared library is present. If that ever changes, it means we'll catch the error earlier as well.

rakuco:
> > The permissions being used to install the files are a bit mixed up, but apparently upstream has the same issue :/
> 
> Yes, I tried to use the same permission that the upstream install script uses. I hope I got them right...

I've sent https://chromium-review.googlesource.com/c/chromium/src/+/4296549 to check if this is just an oversight.

rwmacleod:
Sorry for the delay. I see that this is merged upstream but I've kicked off a build for qemux86-64 qemuarm rpi4-64 and I'll test runtime for at least 2 of those targets.

meta-chromium/recipes-browser/chromium/chromium-gn.inc
```diff
@@ -468,19 +468,33 @@ do_install() {
...
+		install -m 0644 libEGL.so ${D}${libdir}/chromium/
 	fi
 	if [ -e libGLESv2.so ]; then
-		install -m 0755 libGLESv2.so ${D}${libdir}/chromium/
+		install -m 0644 libGLESv2.so ${D}${libdir}/chromium/
+	fi
+
+	# libvulkan (for ANGLE)
+	if [ -e libvulkan.so.1 ]; then
+		install -m 0755 libvulkan.so.1 ${D}${libdir}/chromium/
```
rakuco, line 488: My upstream CL got merged, so I think this can be installed with 0644 as well.

rwmacleod:
Hey, So chromium works fine on master with these changes. 
I've been a bit busy so let me know if I've missed something.

raspberrypi4-64:~$ ls -l /proc/1242/exe 
lrwxrwxrwx    1 foo      foo              0 Mar  7 00:00 /proc/1242/exe -> /usr/lib/chromium/chromium-bin

# looks like we need to change the perms of libvulkan but the EGL, GLES are as expected:
raspberrypi4-64:~$ ls -l /usr/lib/chromium/
-rwsr-xr-x    1 root     root         15480 Mar  9  2018 chrome-sandbox
-rw-r--r--    1 root     root        610313 Mar  9  2018 chrome_100_percent.pak
-rw-r--r--    1 root     root        913218 Mar  9  2018 chrome_200_percent.pak
-rwxr-xr-x    1 root     root        990392 Mar  9  2018 chrome_crashpad_handler
-rwxr-xr-x    1 root     root     215323216 Mar  9  2018 chromium-bin
-rwxr-xr-x    1 root     root          1941 Mar  9  2018 chromium-wrapper
-rw-r--r--    1 root     root      10541264 Mar  9  2018 icudtl.dat
-rw-r--r--    1 root     root         49760 Mar  9  2018 libEGL.so
-rw-r--r--    1 root     root       5767344 Mar  9  2018 libGLESv2.so
-rw-r--r--    1 root     root      19505296 Mar  9  2018 libvk_swiftshader.so
-rwxr-xr-x    1 root     root        422856 Mar  9  2018 libvulkan.so.1
drwxr-xr-x    2 root     root          4096 Mar  9  2018 locales
-rw-r--r--    1 root     root       7574790 Mar  9  2018 resources.pak
-rw-r--r--    1 root     root        162336 Mar  9  2018 snapshot_blob.bin
-rw-r--r--    1 root     root        477936 Mar  9  2018 v8_context_snapshot.bin
-rw-r--r--    1 root     root           107 Mar  9  2018 vk_swiftshader_icd.json

These libraries aren't used by the main process it seems:
raspberrypi4-64:~$ grep GL /proc/1242/maps
7fb01a0000-7fb0217000 r-xp 00000000 b3:02 3985                           /usr/lib/libGL.so.1.2.0
7fb0217000-7fb022d000 ---p 00077000 b3:02 3985                           /usr/lib/libGL.so.1.2.0
7fb022d000-7fb0230000 r--p 0007d000 b3:02 3985                           /usr/lib/libGL.so.1.2.0
7fb0230000-7fb0231000 rw-p 00080000 b3:02 3985                           /usr/lib/libGL.so.1.2.0

but are used by one of the child processes:

raspberrypi4-64:~$ grep GLES /proc/*/maps 2> /dev/null
/proc/1346/maps:7fadb70000-7fadcc1000 r--p 00000000 b3:02 3674                           /usr/lib/chromium/libGLESv2.so
/proc/1346/maps:7fadcc1000-7fadcd0000 ---p 00151000 b3:02 3674                           /usr/lib/chromium/libGLESv2.so
/proc/1346/maps:7fadcd0000-7fae09a000 r-xp 00150000 b3:02 3674                           /usr/lib/chromium/libGLESv2.so
/proc/1346/maps:7fae09a000-7fae0a9000 ---p 0052a000 b3:02 3674                           /usr/lib/chromium/libGLESv2.so
/proc/1346/maps:7fae0a9000-7fae0e0000 r--p 00519000 b3:02 3674                           /usr/lib/chromium/libGLESv2.so
/proc/1346/maps:7fae0e0000-7fae0ef000 ---p 00570000 b3:02 3674                           /usr/lib/chromium/libGLESv2.so
/proc/1346/maps:7fae0ef000-7fae120000 rw-p 0054f000 b3:02 3674                           /usr/lib/chromium/libGLESv2.so
raspberrypi4-64:~$ grep EGL /proc/*/maps 2>/dev/null
/proc/1346/maps:7fadb30000-7fadb35000 r--p 00000000 b3:02 3673                           /usr/lib/chromium/libEGL.so
/proc/1346/maps:7fadb35000-7fadb44000 ---p 00005000 b3:02 3673                           /usr/lib/chromium/libEGL.so
/proc/1346/maps:7fadb44000-7fadb4c000 r-xp 00004000 b3:02 3673                           /usr/lib/chromium/libEGL.so
/proc/1346/maps:7fadb4c000-7fadb5b000 ---p 0001c000 b3:02 3673                           /usr/lib/chromium/libEGL.so
/proc/1346/maps:7fadb5b000-7fadb5c000 r--p 0000b000 b3:02 3673                           /usr/lib/chromium/libEGL.so
/proc/1346/maps:7fadb5c000-7fadb6b000 ---p 0002c000 b3:02 3673                           /usr/lib/chromium/libEGL.so
/proc/1346/maps:7fadb6b000-7fadb6c000 rw-p 0000b000 b3:02 3673                           /usr/lib/chromium/libEGL.so

raspberrypi4-64:~$ cat /proc/1346/cmdline 
/usr/lib/chromium/chromium-bin --type=gpu-process --crashpad-handler-pid=0 --enable-crash-reporter=,stable --change-stack-guard-on-fork=enable --gpu-preferences=WAAAAAAAAAAgAAAIAAAAAAAAAAAAAAAAAABgAAAAAAA4AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAACAAAAAAAAAAIAAAAAAAAAABAAAAAAAAAAgAAAAAAAAACAAAAAAAAAAIAAAAAAAAAA== --use-gl=angle --use-angle=swiftshader-webgl --shared-files --field-trial-handle=0,i,17464429872325917283,7622676165658080321,131072

raspberrypi4-64:~$ cat /proc/1342/cmdline 
/usr/lib/chromium/chromium-bin --type=renderer --crashpad-handler-pid=0 --enable-crash-reporter=,stable --change-stack-guard-on-fork=enable --first-renderer-process --lang=en-US --num-raster-threads=2 --enable-main-frame-before-activation --renderer-client-id=6 --time-ticks-at-unix-epoch=-1678147077178358 --launch-time-ticks=161391578 --shared-files=v8_context_snapshot_data:100 --field-trial-handle=0,i,17464429872325917283,7622676165658080321,131072

That's about all I have time for today. I hope it helps.

meta-chromium/recipes-browser/chromium/chromium-gn.inc
otavio: Using single lines for those copy commands makes it easier to read. For example:

```bash
# Swiftshader VK
[ -e libvk_swiftshader.so ] && install -m 0644 libvk_swiftshader.so ${D}${libdir}/chromium/  
```

meta-chromium/recipes-browser/chromium/chromium-gn.inc
MaxIhlenfeldt: Understood, I've undone this change and modified the ANGLE/Swiftshader to also end with periods.

meta-chromium/recipes-browser/chromium/chromium-gn.inc
MaxIhlenfeldt: I've changed it to match upstream. As https://source.chromium.org/chromium/chromium/src/+/main:chrome/installer/linux/common/installer.include;l=220;drc=01dfcac85a4479e8474e5445cdad6c4997e6b1b8 also just checks for `libEGL.so` for ANGLE and then assumes that `libGLESv2.so` will also be present, I've changed that as well.

Regarding the one-line approach, for me personally I don't think it makes a difference for readability, wdyt @rakuco? If we change this, I think we can do a separate PR that updates all similar `install`s in one go (ANGLE + Swiftshader + `libminigbm.so`).

meta-chromium/recipes-browser/chromium/chromium-gn.inc
MaxIhlenfeldt: Done, thanks for checking this and fixing it upstream!

MaxIhlenfeldt:
Sorry for taking so long to reply, last week has been busy.

> That's about all I have time for today. I hope it helps.

Thanks @rwmacleod, your findings sound reasonable to me, so I think we can ship this.

meta-chromium/recipes-browser/chromium/chromium-gn.inc
rakuco: > Regarding the one-line approach, for me personally I don't think it makes a difference for readability, wdyt @rakuco? If we change this, I think we can do a separate PR that updates all similar installs in one go (ANGLE + Swiftshader + libminigbm.so).

I don't have a strong preference here either. I think I originally wrote these blocks like this because they looked like C++ the most :-) No objections to changing them to one-liners if you prefer, but I agree this can be done separately.

rakuco (approved):
Still LGTM. I think it's better to squash the two commits together, but this looks good to go!

otavio (approved):

MaxIhlenfeldt:
Also, thank you all for the reviews and testing!

otavio:
@MaxIhlenfeldt I enabled the squash option.

MaxIhlenfeldt:
> I enabled the squash option.

Thanks!

PR #698, opened 2023-03-01
chromium: Reduce minimum browser window width to 480px

marex:
The TST043015CNHX panel is a 480x272 panel. Chromium browser limits the minimum window width to 500px since commits:

6b0ecc1e4532c ("Set lower bound on contents pane size.") 580d6900dc7fc ("In tab dragging, the window size should be larger than its minimum size.")

To make chromium browser usable on those tiny panels as well, reduce the minimum window width to 480px. The upstreaming of this change is in progress.

### DIFF
--- a/meta-chromium/recipes-browser/chromium/chromium-gn.inc
+++ b/meta-chromium/recipes-browser/chromium/chromium-gn.inc
@@ -23,6 +23,7 @@ SRC_URI += " \
     file://0013-Fix-html_minifier-script-for-node-v12.patch \
     file://0015-drop-gsimple-template-names.patch \
     file://0016-cast-to-bool-to-avoid-constexpr-error.patch \
+    file://0017-Reduce-minimum-browser-window-width-to-480px.patch \
 "
 
 SRC_URI:append:libc-musl = "\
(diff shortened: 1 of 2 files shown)


rakuco:
> The upstreaming of this change is in progress.

Is there a CL in Gerrit or an upstream bug report you can refer to? I'm weary of taking this patch in without more information, as the new size also looks somewhat arbitrary from a meta-chromium perspective.

marex:
> > The upstreaming of this change is in progress.
> 
> Is there a CL in Gerrit or an upstream bug report you can refer to? I'm weary of taking this patch in without more information, as the new size also looks somewhat arbitrary from a meta-chromium perspective.

@rakuco I sent the patch to the authors of the aforementioned two commits via email as the chromium contribution guidelines are difficult for such a one-liner . I cannot tell whether they created a crbug from that, sorry, but this one https://bugs.chromium.org/p/chromium/issues/detail?id=875197 is very much the same issue (and probably any other crbug which mentions "500px" in some width context)

otavio:
I support the addition of a workaround for this, but we need a long-term vision to remove the workaround. @rakuco, can you fix this upstream?

marex:
@otavio @rakuco has there been any progress ? The patch still applies to latest chromium version too.

otavio:
I support adding this patch. @rakuco ?

arniekanwal:
I realize this PR was opened over a year ago, but I would also like to know if there is a plan to support this patch?

I am building a BSP for a touchscreen device where the configured fullscreen state is 480x800px. Chromium in kiosk/fullscreen mode would fail to launch until I applied this patch. I had similar errors as this bug: #733.

marex:
@kraj @otavio how can we proceed here ?

Issue #714, opened 2023-05-19
yocto-check-layer gives compliance error for meta-chromium layer


nrpt-m:
As explained in:
    https://github.com/OSSystems/meta-browser/commit/2f5fe79935e1f5350f9e487625343a8f27b165d3
Chromium doesn't allow building with system libwayland anymore, and needs (a static version of) libffi to build its own libwayland.

Chromium-ozone-wayland build is looking for libffi_pic.a (static library) at linker stage therefore, a libffi_%.bbappend in the meta-chromium layer was added. Yocto prefers to use shared libraries while building and preferred solution for patching deployed software.

> //Also copy the PIC version of libffi.a to the sysroot, as Chromium needs it.
> // This is the same approach that Debian takes, which Chromium uses for its
> // sysroot: https://sources.debian.org/src/libffi/3.4.4-1/debian/rules/#L103
> do_install:append() {
>     install -m 0644 ${B}/.libs/libffi_convenience.a ${D}/${libdir}/libffi_pic.a
> }

This above command in libffi_%.bbappend causes yocto-check-layer compliance errors. If we move this command to main libffi recipe then, yocto-check-layer errors goes away. As Yocto prefers to use shared library and to resolve the yocto-check-layer compliance errors, we need to fix the build of chromium-ozone-wayand to not require libffi_pic.a ( a static library ).

Upstream References:
https://chromium.googlesource.com/chromium/src/+/lkgr/build/config/linux/libffi/BUILD.gn
https://chromium.googlesource.com/chromium/src/+/188513754c7ff05e7775e6cb2621ee006ebdbf6c

Can we try the below solution to fix the build of chromium-ozone-wayand to not require libffi_pic.a ( a static library ) :
**Source file: chromium-112.0.5615.165/build/config/linux/libffi/BUILD.gn**
In BUILD.gn file, make the variable use_system_libffi always "True" so that, build will always look/use the shared libffi library instead of static libffi_pic.a library and apply this as a patch to check if chromium builds fine with the shared libffi library itself.

rwmacleod:
@nrpt-m  Yes, I think this makes sense since the original issue that required libffi_pic.a was version skew of libffi.so and
the Yocto environment should, by design, avoid that. If  @MaxIhlenfeldt  agrees, @nrpt-m  will send a patch?

[...]

PR #715, opened 2023-05-25
chromium: Make yocto-check-layer pass

nrpt-m:
As explained in:
https://github.com/OSSystems/meta-browser/commit/2f5fe79935e1f5350f9e487625343a8f27b165d3 
Chromium doesn't allow building with system libwayland anymore, and needs (a static version of) libffi to build its own libwayland. Therefore, libffi_%.bbappend was added to make libffi_pic.a (static library) available while building chromium-ozone-wayland.

This libffi_%.bbappend is causing yocto-check-layer to give compliance error for meta-chromium layer.

By adding "use_system_libffi=true" to GN_ARGS in chromium-ozone-wayland recipe and removing the libffi_%.bbappend, we can make chromium-ozone-wayland to build with libffi.so (shared library). With these changes, we can avoid using libffi_pic.a (static library) for chromium-ozone-wayland build and also resolve yocto-check-layer compliance error for meta-chromium layer.

Added Upstream-Status to required patches to fix test_patches_upstream_status failure in yocto-check-layer.

Fixes: #714 

Signed-off-by: Narpat Mali <narpat.mali@windriver.com>

### DIFF
--- a/meta-chromium/recipes-browser/chromium/files/musl/no-res-ninit-nclose.patch
+++ b/meta-chromium/recipes-browser/chromium/files/musl/no-res-ninit-nclose.patch
@@ -1,6 +1,8 @@
 Source: https://git.alpinelinux.org/aports/plain/community/chromium/no-res-ninit-nclose.patch
 similar to dns-resolver.patch, musl doesn't have res_ninit and so on
 
+Upstream-Status: Pending
+
 Signed-off-by: Khem Raj <raj.khem@gmail.com>
 --- a/net/dns/public/scoped_res_state.cc
 +++ b/net/dns/public/scoped_res_state.cc
--- a/meta-chromium/recipes-browser/chromium/chromium-ozone-wayland_112.0.5615.165.bb
+++ b/meta-chromium/recipes-browser/chromium/chromium-ozone-wayland_112.0.5615.165.bb
@@ -21,6 +21,7 @@ GN_ARGS += "\
         use_xkbcommon=true \
         use_system_minigbm=true \
         use_system_libdrm=true \
+        use_system_libffi=true \
         use_gtk=false \
 "
 
--- a/meta-chromium/recipes-browser/chromium/files/0009-nomerge-attribute-on-declaration-is-only-available-s.patch
+++ b/meta-chromium/recipes-browser/chromium/files/0009-nomerge-attribute-on-declaration-is-only-available-s.patch
@@ -20,6 +20,8 @@ build error: 'nomerge' attribute cannot be applied to a declaration
 See https://reviews.llvm.org/D92800
 
 Change-Id: I32e1f7dc9049737d54d8a16de5308aa5aae1ced1
+
+Upstream-Status: Inappropriate
 ---
  base/compiler_specific.h | 3 ++-
  1 file changed, 2 insertions(+), 1 deletion(-)
--- a/meta-chromium/recipes-browser/chromium/files/0017-Revert-Fix-std-span-autodetection-7805.patch
+++ b/meta-chromium/recipes-browser/chromium/files/0017-Revert-Fix-std-span-autodetection-7805.patch
@@ -4,6 +4,8 @@ Date: Tue, 11 Apr 2023 15:37:58 +0200
 Subject: [PATCH] Revert "Fix std::span autodetection (#7805)"
 
 This reverts commit a6f41944899e65b38ec45ee82f6840248f06b471.
+
+Upstream-Status: Inappropriate
 ---
  .../src/include/flatbuffers/stl_emulation.h | 19 ++++++++-----------
  1 file changed, 8 insertions(+), 11 deletions(-)
--- a/meta-chromium/recipes-browser/chromium/files/0018-Only-default-operator-on-declaration.patch
+++ b/meta-chromium/recipes-browser/chromium/files/0018-Only-default-operator-on-declaration.patch
@@ -8,7 +8,8 @@ the first declaration was removed in P2085R0, which clang only supports
 in versions 14 and above. Because dunfell still uses clang 12, we need
 to fix the few occurrences where this requirement isn't met.
 
-Upstream-Status: Unknown
+Upstream-Status: Inappropriate
+
 Signed-off-by: Max Ihlenfeldt <max@igalia.com>
 ---
  .../core/browser/affiliation/affiliation_fetcher_base.cc     | 3 ---
--- a/meta-chromium/recipes-browser/chromium/libffi_%.bbappend
+++ b/meta-chromium/recipes-browser/chromium/libffi_%.bbappend
@@ -1,6 +1,0 @@
-# Also copy the PIC version of libffi.a to the sysroot, as Chromium needs it.
-# This is the same approach that Debian takes, which Chromium uses for its
-# sysroot: https://sources.debian.org/src/libffi/3.4.4-1/debian/rules/#L103
-do_install:append() {
-	install -m 0644 ${B}/.libs/libffi_convenience.a ${D}/${libdir}/libffi_pic.a
-}


rakuco (commented):
Looks fine to me, although I need to defer testing to @MaxIhlenfeldt. Thanks for working on this!

From a process perspective, could you add something like [Fixes #714](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword) to the PR and commit message to connect this PR to the corresponding issue?

nrpt-m:
> Looks fine to me, although I need to defer testing to @MaxIhlenfeldt. Thanks for working on this!
> 
> From a process perspective, could you add something like [Fixes #714](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword) to the PR and commit message to connect this PR to the corresponding issue?

Added.

MaxIhlenfeldt:
Thanks for the PR!

Unfortunately, I don't really have a good runtime testing setup. @rwmacleod I think you usually test chromium-x11, would it be possible for you to test chromium-wayland in this case?

MaxIhlenfeldt:
> /buildarea/eng1/wrls/lts23/layers/meta-browser/meta-chromium/recipes-browser/chromium/files/0009-nomerge-attribute-on-declaration-is-only-available-s.patch

I think this patch shouldn't be needed anymore. It looks like it's only needed when building with clang < 12, which even on dunfell isn't the case anymore. I'll check if it can be removed as part of the 113 update PR.

> /buildarea/eng1/wrls/lts23/layers/meta-browser/meta-chromium/recipes-browser/chromium/files/0017-Revert-Fix-std-span-autodetection-7805.patch

Ah, I wanted to check with the relevant upstream maintainers regarding this one as part of the update to 112 but forgot, sorry about that. I still don't know why it's needed in the first place, but maybe they're still willing to accept the patch upstream. Feel free to ask them if you want, or I can do it, whatever works best for you.

> /buildarea/eng1/wrls/lts23/layers/meta-browser/meta-chromium/recipes-browser/chromium/files/0018-Only-default-operator-on-declaration.patch

I don't think they accept patches that are needed for building with clang 12, as that's not supported (we have some other clang 12-specific patches marked as Inappropriate).

kraj (approved):
Generally I agree with this patch, however, chromium vendors all its dependencies and this might become problematic over time, although this library is not changed as frequently so we might be ok. It would be good to ask upstream if they intend to keep supporting using system provided libffi in future.

nrpt-m:
> Will test chromium-ozone-wayland for kirkstone & dunfell branches tomorrow.

Have tested chromium-ozone-wayland for kirkstone & dunfell branches and below are my observations:

1. For master & kirkstone branch, core-image-weston image with chromium-ozone-wayland builds successfully and we see that the default id comes something like this: "**uid=1000(weston) gid=1000 (weston) groups=19 (input), 44(video) 1000 (weston)**". With "chromium" command chromium launches, works fine and uses system libffi.so.8 library.

> sh-5.28 ldd /usr/lib/chromium/chromium-bin | grep ffi 
> libffi.so.8 => /usr/lib/libffi.so.8 (0x00007f8938895000)

2. For dunfell branch, core-image-weston image with chromium-ozone-wayland builds successfully and but, we see that the default id comes something like this: "**uid=0(root) gid=0(root) groups=0(root)**". With "chromium" command chromium doesn't launches and it asks us use "chromium --no-sandbox" and with this command chromium launches, works fine and uses system libffi.so.7 library. If I switch to some user and run chromium then it says permission denied -> have attached a screenshot of it.

> sh-5.28 ldd /usr/lib/chromium/chromium-bin | grep ffi
> libffi.so.7=> /usr/lib/libffi.so.7 (0x00007ff7fdac8000)
chromium-ozone-wayland-dunfell

kraj:
> > Will test chromium-ozone-wayland for kirkstone & dunfell branches tomorrow.
> 
> Have tested chromium-ozone-wayland for kirkstone & dunfell branches and below are my observations:
> 
> 1. For master & kirkstone branch, core-image-weston image with chromium-ozone-wayland builds successfully and we see that the default id comes something like this: "**uid=1000(weston) gid=1000 (weston) groups=19 (input), 44(video) 1000 (weston)**". With "chromium" command chromium launches, works fine and uses system libffi.so.8 library.
> 
> > sh-5.28 ldd /usr/lib/chromium/chromium-bin | grep ffi
> > libffi.so.8 => /usr/lib/libffi.so.8 (0x00007f8938895000)
> 
> 2. For dunfell branch, core-image-weston image with chromium-ozone-wayland builds successfully and but, we see that the default id comes something like this: "**uid=0(root) gid=0(root) groups=0(root)**". With "chromium" command chromium doesn't launches and it asks us use "chromium --no-sandbox" and with this command chromium launches, works fine and uses system libffi.so.7 library. If I switch to some user and run chromium then it says permission denied -> have attached a screenshot of it.
> 
> > sh-5.28 ldd /usr/lib/chromium/chromium-bin | grep ffi
> > libffi.so.7=> /usr/lib/libffi.so.7 (0x00007ff7fdac8000)
> 
> chromium-ozone-wayland-dunfell

I think thats expected. in dunfell default user for weston shell was root where as in kirkstone it was changed to weston ( non-root user )

nrpt-m:
@kraj, have added upstream-status for some of the patches so, could you please review it and let me know if these are correct or not.

meta-chromium/recipes-browser/chromium/files/musl/no-res-ninit-nclose.patch
```diff
@@ -1,6 +1,8 @@
 Source: https://git.alpinelinux.org/aports/plain/community/chromium/no-res-ninit-nclose.patch
 similar to dns-resolver.patch, musl doesn't have res_ninit and so on
 
+Upstream-Status: Backport
```
kraj, line 4: also add the link to upstream commit here 
Upstream-Status: Backport [link]

nrpt-m: Added.

meta-chromium/recipes-browser/chromium/files/musl/no-res-ninit-nclose.patch
```diff
@@ -1,6 +1,7 @@
-Source: https://git.alpinelinux.org/aports/plain/community/chromium/no-res-ninit-nclose.patch
 similar to dns-resolver.patch, musl doesn't have res_ninit and so on
 
+Upstream-Status: Backport [https://git.alpinelinux.org/aports/plain/community/chromium/no-res-ninit-nclose.patch]
```
kraj, line 3: but this is not a backport. Backport should point to location where it got applied to chromium upstream. In this case perhaps status should be Pending and mention source from alpine location.

nrpt-m: Sure, done.

rakuco (commented):
The changes look fine to me, but I think it'd be better to have separate commits in this PR for the libffi part and the Upstream-Status bits since they are doing independent things. The PR title can then be updated to something more generic like "chromium: Make yocto-check-layer pass".

rakuco:
> Generally I agree with this patch, however, chromium vendors all its dependencies and this might become problematic over time, although this library is not changed as frequently so we might be ok. It would be good to ask upstream if they intend to keep supporting using system provided libffi in future.

Looking at https://chromium.googlesource.com/chromium/src/+/main/build/config/linux/libffi/BUILD.gn, this should be safe as long as ChromeOS continues to use libffi as a shared library. According to the comment there, they are basically trying to avoid a mismatch between the version used to build Chromium and the version used at runtime. I believe OE/Yocto aren't prone to this because a change to libffi should cause Chromium to be rebuilt as well (IIRC).

nrpt-m:
> The changes look fine to me, but I think it'd be better to have separate commits in this PR for the libffi part and the Upstream-Status bits since they are doing independent things. The PR title can then be updated to something more generic like "chromium: Make yocto-check-layer pass".

Done.

rakuco (approved):
lgtm!

Issue #723, opened 2023-06-06
chromium: Prepare for Rust.


rakuco:
This is kind of a meta-bug to make everyone aware of https://groups.google.com/a/chromium.org/g/chromium-packagers/c/Gl30hsQ85g8/m/XTc46sRgAQAJ -- tl;dr at some point in the not-too-distant future building Chromium will also require a Rust toolchain.

I know basically nothing about Rust and Rust toolchains and I can imagine integrating one into our build will not be fun. And supporting dunfell will probably make things extra challenging :-)

otavio:
Dunfell will be challenging, mainly due to the difference in how it fetches crates. In Dunfell's case, meta-rust could be an option; however, we will need to cut it off eventually.

rwmacleod:
We'll have to see what version of Rust is required to decide if we keep dunfell going. 

For kirkstone, we can also consider using:
https://git.yoctoproject.org/meta-lts-mixins/log/?h=kirkstone/rust-1.68

In theory that layer can be used to provide the same ( or closer to the same ) version of Rust used on master. We could make a mickledore branch as well. Anyway, it seems like we can only wait for now.

rwmacleod:
> I think once we have M113 there should be enough code for one to experiment passing `enable_rust=true` to GN (it's actually true by default in more recent milestones), seeing what's required and what parts of the toolchain stuff mentioned in the chromium-packagers email needs to be replicated without Chromium's bundled toolchain.

Unfortunately there's a new zero-day high severity CVE that is fixed in 114 so unless Rust is required for 114, we should do that update and then work on Rust support. 

https://hothardware.com/news/google-emergency-0-day-patch-billions-chrome-installs

MaxIhlenfeldt:
Looks like m115 is the first version that needs a Rust toolchain:
```
| ERROR at //build/config/rust.gni:126:24: Script returned non-zero exit code.
|       rustc_revision = exec_script("//tools/rust/update_rust.py",
|                        ^----------
| Current dir: /home/gitlab-runner/yocto-chromium/builds/ci-chromium-wayland-qemuarm64-mickledore/tmp-glibc/work/cortexa57-oe-linux/chromium-ozone-wayland/115.0.5790.98-r0/chromium-115.0.5790.98/out/Release/
[...]

PR #741, opened 2023-08-10
chromium: update to 115.0.5790.170

MaxIhlenfeldt:
Release notes:
    https://chromereleases.googleblog.com/2023/07/stable-channel-update-for-desktop.html
    https://chromereleases.googleblog.com/2023/07/stable-channel-update-for-desktop_20.html
    https://chromereleases.googleblog.com/2023/07/stable-channel-update-for-desktop_25.html
    https://chromereleases.googleblog.com/2023/08/stable-channel-update-for-desktop.html

Build and patch changes:
------------------------

Drop 0019-Backport-constexpr-std-string-fixes.patch,
0022-Backport-Avoid-capturing-structured-bindings.patch,
0026-Backport-std-string-ends_with-is-C-20.-Use-base-Ends.patch, and
0030-Backport-Avoid-operator-which-is-C-20-specific.patch as they are
included upstream now.

Add one more fix to 0018-Only-default-operator-on-declaration.patch,
which previously was part of
0030-Backport-Avoid-operator-which-is-C-20-specific.patch to prevent a
conflict. Also drop some fixes that are now included upstream.

Drop some fixes from 0021-Add-missing-typename-s.patch that are now
included upstream, and add two new ones.

Add one more fix to 0028-Avoid-capturing-structured-bindings.patch.

Add 0033-Revert-Reland-Reland-mte-refactor-the-tagging-functi.patch to
work around a bug in clang 14 that leads to a build error.

Add 0034-Avoid-parenthesized-initialization-of-aggregates.patch to fix
build errors with clang < 16.

Add 0035-Backport-IWYU-for-chromedriver-chrome-client_hints.h.patch to
fix a build error.

Rebase remaining patches.

License changes:
----------------

Added licenses:
  * third_party/beto-core/LICENSE
  - https://chromium-review.googlesource.com/c/chromium/src/+/4444306

  * third_party/devtools-frontend/src/front_end/third_party/puppeteer-replay/package/LICENSE
  - https://chromium-review.googlesource.com/c/devtools/devtools-frontend/+/4480871

Removed licenses: none.

Updated licenses: none.

Test-built:
-----------

* chromium-ozone-wayland:
 - mickledore, clang, MACHINE=qemuarm64
 - kirkstone, clang,  MACHINE=raspberrypi4-64

* chromium-x11
 - dunfell, clang**,  MACHINE=raspberrypi4-64

** Please note that Chromium requires below set-up when on dunfell
branch.
- The clang version to be >= 14 and for that, use the latest
[...]

PR #745, opened 2023-08-22
chromium: Disable Rust for now

MaxIhlenfeldt:
It turns out that Rust might be used in the build in 114 already. As discussed in #723, we need to figure out how to best support Rust builds, and until we've done so we need to disable it.

### DIFF
--- a/meta-chromium/recipes-browser/chromium/chromium-gn.inc
+++ b/meta-chromium/recipes-browser/chromium/chromium-gn.inc
@@ -303,6 +303,9 @@ GN_ARGS += ' \
 # but we can use whatever user configured in PARALLEL_MAKE
 GN_ARGS += 'max_jobs_per_link="${@oe.utils.parallel_make_argument(d, '%d')}"'
 
+# We haven't figured out how to best support Rust yet, so disable it for now.
+GN_ARGS += 'enable_rust=false'
+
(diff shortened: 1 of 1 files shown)


MaxIhlenfeldt:
Given that this change was already part of #741, which has been extensively tested, I suppose we can skip the tests for this PR?

meta-chromium/recipes-browser/chromium/chromium-gn.inc
```diff
@@ -303,6 +303,9 @@ GN_ARGS += ' \
 # but we can use whatever user configured in PARALLEL_MAKE
 GN_ARGS += 'max_jobs_per_link="${@oe.utils.parallel_make_argument(d, '%d')}"'
 
+# We haven't figured out how to best support Rust yet, so disable it for now.
+GN_ARGS += 'enable_rust=false'
```
rakuco, line 307: Please refer to #723 in the comment too.

rakuco (approved):
lgtm, I agree we can skip the extensive testing for this PR.

meta-chromium/recipes-browser/chromium/chromium-gn.inc
MaxIhlenfeldt: Done

PR #751, opened 2023-09-09
chromium-gn.inc: make alsa configurable

vermaete:
based on the DISTRO_FEATURE alsa


otavio (approved):

rakuco:
The patch looks sensible, but I'd like to confirm one thing: does a lack of "alsa" in `DISTRO_FEATURES` mean the target system has no sound or something like that, or is the idea to use a different backend?

vermaete:
Hi,

It was just a matter of having my image slightly smaller for my product
without a need for sound.

Br,


Op ma 11 sep. 2023 om 09:16 schreef Raphael Kubo da Costa <
***@***.***>

rakuco:
Got it, thanks. Since the `DISTRO_FEATURE` entry is already there, I think it doesn't hurt to do this.

rakuco (approved):

PR #753, opened 2023-09-21
chromium: Require clang 14.0.0 in do_check_llvm_version()

rakuco:
We have required clang 14 since the update to Chromium 115, so make sure the version check reflects this and points people to the right branch.


kraj (approved):
lgtm

PR #758, opened 2023-10-02
chromium: update to 116.0.5845.187

MaxIhlenfeldt:
Release notes:
    https://chromereleases.googleblog.com/2023/09/stable-channel-update-for-desktop_11.html

Build and patch changes:
------------------------

None.

License changes:
----------------

None.

Test-built:
-----------

* chromium-ozone-wayland:
 - mickledore, clang, MACHINE=qemuarm64
 - kirkstone, clang,  MACHINE=raspberrypi4-64

* chromium-x11
 - dunfell, clang**,  MACHINE=raspberrypi4-64

** Please note that Chromium requires below set-up when on dunfell branch.
 - The clang version to be >= 14 and for that, use the latest meta-clang/dunfell-clang14 branch.
 - Require the latest meta-oe with Nodejs 14.x support.
 - Add the PREFERRED_VERSION_nodejs-native = "14.%" in conf/local.conf file.


MaxIhlenfeldt:
@rakuco, as this is just a patch update, I don't think we need to do rigorous testing like we do for major version updates. I'm currently building the three listed variants to make sure no build errors sneak in, but I think as long as they all build we should be fine.

rakuco (approved):
lgtm

rakuco:
> @rakuco, as this is just a patch update, I don't think we need to do rigorous testing like we do for major version updates. I'm currently building the three listed variants to make sure no build errors sneak in, but I think as long as they all build we should be fine.

Sounds fine to me. Thanks for working on this!

imoralessirgo:
Hello :) thanks again for working on this. Just wanted to check-in and ask what the expected timeline for this patch landing is?

MaxIhlenfeldt:
Thanks for the reminder - all builds passed, so this should be fine!

PR #762, opened 2023-10-15
chromium: Fix SIGILL on RPI4

kraj:


rwmacleod:
@kraj  Seems sensible. Have you tested this enough or do we need @nrpt-m  to generate a new RPi4 test image?

kraj:
I have tested rpi4 for both glibc and musl

meta-chromium/recipes-browser/chromium/files/boringssl_no_asm_config.patch
```diff
@@ -0,0 +1,18 @@
+Disable assembly routines on arm64/rpi4 since it uses
+AES crypto extentions and these extentions do not exist in rpi4
+
+This fixes random SIGILL seen in aes_hw_set_encrypt_key() on rpi4-64
+
+Upstream-Status: Pending
```
rakuco, line 6: I think the upstream-status value here should be "Inappropriate", I doubt upstream will be interested in such a patch and it's fixing a crash on systems that are not officially supported upstream AFAICS.

rakuco (approved):
lgtm with a nitpick

meta-chromium/recipes-browser/chromium/files/boringssl_no_asm_config.patch
kraj: I thought it could atleast start a convesation and let chromium devs state that, crypto is much have for armv8 for chromium thats why pending

rakuco: I see. In this case it's the BoringSSL community (https://boringssl.googlesource.com/boringssl) more than the Chromium folks. If you plan on reaching out to them, then I have no objection to using "Pending" (with a link to a bug, discussion or patch) here.

Still lgtm either way.

kraj: Let’s merge it for
Now I will keep it going with chromium community I think the issue of how gn is invoked on it by chromium

PR #782, opened 2024-01-22
chromium: Enable Rust

MaxIhlenfeldt:
This adds a dependency on OE's Rust recipe, as well as all the necessary changes to make Chromium successfully use that toolchain.

This change will also mean we don't support any Yocto release older than kirkstone, as that is the first release with Rust in OE that isn't EOL yet. Most notably, we will drop support for dunfell.

### DIFF
--- a/meta-chromium/recipes-browser/chromium/chromium-gn.inc
+++ b/meta-chromium/recipes-browser/chromium/chromium-gn.inc
@@ -45,6 +45,9 @@ SRC_URI += "\
     file://0020-Fix-implicitly-deleted-default-constructor-build-err.patch \
     file://0021-Don-t-delete-CrashKeyWithName-dtor.patch \
     file://0022-Use-base-ranges-instead-of-std-ranges.patch \
+    file://0023-Use-the-correct-path-to-libclang_rt.builtins.a.patch \
+    file://0024-Adjust-the-Rust-build-to-our-needs.patch \
+    file://0025-Make-toolchain_supports_rust_thin_lto-configurable.patch \
 "
 # ARM/AArch64-specific patches.
 SRC_URI:append:arm = "\
@@ -107,9 +110,11 @@ DEPENDS += " \
     pkgconfig-native \
     ${@bb.utils.contains('DISTRO_FEATURES', 'pulseaudio', 'pulseaudio', '', d)} \
     qemu-native \
+    rust \
+    rust-native \
     virtual/libgl \
 "
-DEPEND:append:runtime-llvm = " compiler-rt-native libcxx-native"
+DEPENDS:append:runtime-llvm = " compiler-rt compiler-rt-native libcxx-native"
 DEPENDS:append:libc-musl = " libexecinfo"
 
 LDFLAGS:append:libc-musl = " -lexecinfo"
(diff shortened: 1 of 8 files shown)


MaxIhlenfeldt:
Some comments:
- @nrpt-m as always, testing would be greatly appreciated! As discussed, this patch will mean we don't support dunfell any more. Also, please use https://git.yoctoproject.org/meta-lts-mixins/log/?h=kirkstone/rust-1.68 for kirkstone. I suspect the build will still fail even with that mixin because the Rust version is too old; but we should confirm it before adding a mixin with a newer version.
- iiuc, one usually doesn't depend on `rust(-native)` directly, but inherits from the `cargo` class. I chose not to do that because it created some problems (I'm not 100% sure what it was, but I think the `libstd-rs` libraries were still being installed despite the line in `rust_%.bbappend`). cc @rwmacleod (I think you are/were involved in meta-rust development?)
- As soon as I get confirmation from someone else that the build works, I'll submit `0025-Make-toolchain_supports_rust_thin_lto-configurable.patch` upstream. I've talked to danakj@, she seemed open to accepting it.

rwmacleod:
@MaxIhlenfeldt  These days I'm mostly involved in Rust in trying to help @sundeep-kokkonda and @Yashinde145  so I've mentioned this thread to them and I see that Sundeep cloned meta-browser so he may reply to your question.

Also, for kirkstone, @g-scott-murray  mentioned today that he is working on an update to (1.70?) of the rust mixin layer and that they are interested in chromium for AGL as you might already know.

nrpt-m:
> Some comments:
> 
> * @nrpt-m as always, testing would be greatly appreciated! As discussed, this patch will mean we don't support dunfell any more. Also, please use https://git.yoctoproject.org/meta-lts-mixins/log/?h=kirkstone/rust-1.68 for kirkstone. I suspect the build will still fail even with that mixin because the Rust version is too old; but we should confirm it before adding a mixin with a newer version.

@MaxIhlenfeldt ,
Have successfully compiled the current chromium-120.0.6099.224 version after applying the patches mentioned in this PR. Have used poky/kirkstone source and all other required layers in kirkstone branch along with this https://git.yoctoproject.org/meta-lts-mixins/log/?h=kirkstone/rust-1.68 layer.

Have attached the console.log file as well.
[console-latest.log](https://github.com/OSSystems/meta-browser/files/14027908/console-latest.log)

MaxIhlenfeldt:
Great to hear it works with the 1.68 mixin on kirkstone already!

MaxIhlenfeldt:
One note, as Rust is a hard requirement for Chromium 121, it'd be easiest to have this PR merged before I start working on the update. I'd also like to do a separate PR that removes all dunfell-specific patches first. But if it takes too long to merge this, I can start working on the update already.

MaxIhlenfeldt:
The `toolchain_supports_rust_thin_lto` patch has been accepted upstream as https://crrev.com/c/5233602, so I moved the patch file to `files/backports` and updated its `Upstream-Status`.

MaxIhlenfeldt:
@kraj @rakuco gentle ping as merging this before starting to work on the (already released) 121 update would be best.

otavio (approved):

kraj (approved):
lgtm

meta-chromium/recipes-browser/chromium/chromium-gn.inc
```diff
@@ -107,9 +110,11 @@ DEPENDS += " \
     pkgconfig-native \
     ${@bb.utils.contains('DISTRO_FEATURES', 'pulseaudio', 'pulseaudio', '', d)} \
     qemu-native \
+    rust \
+    rust-native \
     virtual/libgl \
 "
-DEPEND:append:runtime-llvm = " compiler-rt-native libcxx-native"
```
rakuco, line 112: Out of curiosity, does this mean that this line was completely broken before?

rakuco (commented):
Sorry for the delay here. I know very little about Rust and how its toolchain works, so I can't really review the changes here but fully trust you and the others on this one.

Thanks for all the hard work, and I'm glad to see dunfell support finally going away!

rakuco:
> This change will also mean we don't support any Yocto release older than kirkstone, as that is the first release with Rust in OE that isn't EOL yet. Most notably, we will drop support for dunfell.

@otavio @kraj @MaxIhlenfeldt we'd normally create a "dunfell" branch at the commit ea731e8d4fd8a0b40448ea286b2c92824ee591cc; I wonder if it makes sense given dunfell as a whole is going EoL in just a few months.

otavio:
I am not opposed as this is the last working version.

rakuco:
> I am not opposed as this is the last working version.

Done!

meta-chromium/recipes-browser/chromium/chromium-gn.inc
MaxIhlenfeldt: I think it was working before, but when Rust is enabled Chromium needs to link against the target clang runtime libraries, see https://source.chromium.org/chromium/chromium/src/+/main:build/config/compiler/BUILD.gn;l=1660-1668;drc=ab6484854e1fb7a38160336acc6b34b6ecdda628.

MaxIhlenfeldt: Oh, right, I overlooked that I changed that as well. Yes, that might have indeed been the case :sweat_smile:

rakuco: Interesting. That might explain some of the bug reports we've had over the years :-) Glad to see it's been fixed.

PR #788, opened 2024-02-05
chromium: Drop clang 12-specific patches

MaxIhlenfeldt:
Build and patch changes:
------------------------

The following patches were only needed to get clang 12 to build Chromium, which we previously used on dunfell. We recently switched to using clang 14 on dunfell, and have now dropped dunfell support, so they can be removed:
* 0001-limit-number-of-LTO-jobs.patch
* 0007-Revert-Use-ffile-compilation-dir-instead-of-fdebug-c.patch
* 0008-drop-gsimple-template-names.patch
* 0009-cast-to-bool-to-avoid-constexpr-error.patch
* 0010-Revert-Fix-std-span-autodetection-7805.patch
* 0011-Only-default-operator-on-declaration.patch

Renamed the remaining patches to fit the numbering scheme.

License changes:
----------------

None.

Test-built:
-----------

* chromium-ozone-wayland:
 - nanbield, clang,  MACHINE=qemuarm64
 - kirkstone, clang, MACHINE=raspberrypi4-64


MaxIhlenfeldt:
Did the mentioned test builds already, but without the "rename the remaining patches" part (so that I didn't have to fix the numbering again, should it turn out that we still need one or more of the removed patches). To make sure I didn't mess up anything while renaming, one more build is currently running.

rakuco (approved):
big lgtm! so glad to see all these patches going away :tada:

MaxIhlenfeldt:
Build went through, seems I got the renaming right!

PR #801, opened 2024-03-22
chromium: Drop patches that are specific to clang < 17

MaxIhlenfeldt:
Given that we have decided not to support kirkstone until it uses a more recent version of clang, we can now drop all patches that are only needed to fix errors with clang < 17 (as nanbield uses clang 17).

This also updates LAYERSERIES_COMPAT, dropping kirkstone, as well as the EOL'd mickledore and langdale.


MaxIhlenfeldt:
Marking as draft as the final builds with nanbield and master are still running, but I'm fairly confident that both should work.

rakuco (approved):
image

This is beautiful!

MaxIhlenfeldt:
Builds succeeded, merging now!

PR #822, opened 2024-06-13
firefox: README: mention fork with newer firefox

shr-project:


meta-firefox/README.md
```diff
@@ -1,6 +1,14 @@
...
 ================================================
 
+This layer wasn't updated for couple years and firefox here
+is so old that nobody should use it. If you're looking for
+newer firefox recipes, then look at this fork:
+https://github.com/webdino/meta-browser
+I haven't tried it and I have no idea why they never sent
+any PR back, but at least it has newer version and higher
+chance that it will actually build than the firefox-68 here.
+
```
kraj, line 11: I think it would be better to ecourage upstreaming the changes. Or perhaps removing firefox recipes from meta-browser instead of pointing to a fork.

shr-project: This PR (and issue ticket in their fork) is trying to encourage them to upstream the changes, but if they don't then better to at least point there as people are more likely to find this layer in layerindex.

PR #830, opened 2024-08-22
chromium: fix do_copy_clang_library failed for multilib

hongxu-jia:
Due to commit [clang: use nonarch_libdir/clang for all runtime files][1] applied, the clang runtime libraries was installed to ${nonarch_libdir}, no matter the value of libdir

Tweak it for do_copy_clang_library, use ${nonarch_libdir} to instead

[1] https://github.com/kraj/meta-clang/commit/cda1376a97b86d6ae5c92fd931f9583430c73885


kraj (approved):
lgtm

Issue #849, opened 2024-11-18
chromium-ozone-wayland: eglCreateContext: Requested version is not supported


MastaG:
When I run chromium-ozone-wayland on my armv7 board using the Panfrost mesa drivers, I always get the following error:
```
ERROR:gl_display.cc(497)] EGL Driver message (Error) eglCreateContext: Requested version is not supported
```
The browser runs fine otherwise and seems fully accelerated except for WebGL.
Is it possible I could get more verbose information regarding the above error?
My chromium flags:
`--ozone-platform=wayland --enable-wayland-ime --use-gl=angle --use-angle=gles-egl --use-cmd-decoder=passthrough --ignore-gpu-blocklist --ignore-gpu-blacklist --enable-accelerated-video-decode --enable-features=VaapiVideoDecoder,VaapiVideoEncoder --gtk-version=4 --disable-gpu-memory-buffer-video-frames`

By the way: `--use-gl=egl` doesn't seem to work anymore and needs to be replaced with: `--use-gl=angle --use-angle=gles-egl` or `--use-gl=angle --use-angle=gl-egl`

MastaG:
With `--use-angle=gl-egl` I'm getting the same EGL error.
Is it possible to compile Chromium with legacy EGL support (`--use-gl=egl`)?
Because Angle doesn't seem to work for my platform.

msisov:
> With `--use-angle=gl-egl` I'm getting the same EGL error. Is it possible to compile Chromium with legacy EGL support (`--use-gl=egl`)? Because Angle doesn't seem to work for my platform.

Can you try https://github.com/OSSystems/meta-browser/pull/850 ?

MastaG:
Yes I can only run with Angle now. So I only have `--use-angle=gles-egl`.
When I use `--use-gl=egl`  I'm getting a message that it isn't supported any longer.

Hardware acceleration works for the most part when using Angle.
It's just that it will display the: `ERROR:gl_display.cc(497)] EGL Driver message (Error) eglCreateContext: Requested version is not supported` twice.
And WebGL seems to only thing that fails to work.

See the report:

[about-gpu-2024-11-19T11-15-51-085Z.txt](https://github.com/user-attachments/files/17814344/about-gpu-2024-11-19T11-15-51-085Z.txt)

msisov:
Can you try `--use-angle=gl-egl` ?

MastaG:
Yeah that's probably it.
[...]

PR #850, opened 2024-11-18
chromium: Update EGL CLI args

MaxIhlenfeldt:
As mentioned in e.g. #849, `--use-gl=egl` doesn't work any more. We can replace it with `--use-gl=angle --use-angle=gles-egl`, or just `--use-angle=gles-egl` as ANGLE is the default.


rakuco (approved):
lgtm!

MastaG:
tested and working on both X11 and Wayland.

msisov (approved):

MaxIhlenfeldt:
Thanks for testing!

PR #858, opened 2024-12-16
[scarthgap] chromium: Drop patch to decrease minimum Rust version

### LINKED ISSUE #855: Decide how to handle Scarthgap meta-lts-mixins dependency
meta-lts-mixins now has a `scarthgap/rust` branch: https://git.yoctoproject.org/meta-lts-mixins/log/?h=scarthgap/rust.

To avoid re-introducing #789 for Scarthgap, I think we should declare a dependency on meta-lts-mixins (as soon as we drop the patch to keep the required Rust version at 1.75) - but of course we don't have that dependency for Styhead or Walnascar.

One way to solve this problem would be to stop using the master branch for all currently supported releases, and instead have one branch per Yocto release (which we currently only do once we stop supporting a release). In my experience we've always been in the minority with our current approach - and switching to one branch per release by default would have other advantages as well, e.g. not blocking the update for other releases if the build is only broken for one release.

@rakuco @kraj @otavio wdyt? If we switch our branch approach, how can we best handle it to break as few user's setups as possible?

MaxIhlenfeldt:
Fixes #855.

Build and patch changes:
------------------------

Drop a patch to decrease the required version of Rust, and instead depend on meta-lts-mixins to provide a more up-to-date version of Rust than OE Core.

License changes:
----------------

None.

Test-built:
-----------

* chromium-ozone-wayland:
 - scarthgap, clang, MACHINE=raspberrypi4-64


rakuco (approved):
lgtm

PR #864, opened 2025-01-01
chromium: Add v4l2 m2m stateless decode configuration options

marex:
Add configuration options to enable hardware video decoding using stateless V4L2 M2M device. This allows offloading e.g. h264 video playback to Hantro VPU on i.MX8MP where this was tested. To make that work, enable `use-v4l2` and `proprietary-codecs` PACKAGECONFIG.

For i.MX8MP the following additional udev rules are mandatory:
```
SUBSYSTEM=="video4linux", ATTR{name}=="nxp,imx8mm-vpu-g1-dec", SYMLINK+="video-dec%n"
SUBSYSTEM=="video4linux", ATTR{name}=="nxp,imx8mm-vpu-g2-dec", SYMLINK+="video-dec%n"
SUBSYSTEM=="media", ATTR{model}=="hantro-vpu", SYMLINK+="media-dec%n"
```

This commit was implemented with much great help from Jianfeng Liu .

### DIFF
--- a/meta-chromium/recipes-browser/chromium/chromium-gn.inc
+++ b/meta-chromium/recipes-browser/chromium/chromium-gn.inc
@@ -112,7 +112,7 @@ BUILD_CC:toolchain-clang = "clang"
 BUILD_CXX:toolchain-clang = "clang++"
 BUILD_LD:toolchain-clang = "clang"
 
-PACKAGECONFIG ??= "upower use-egl"
+PACKAGECONFIG ??= "upower use-egl use-v4l2"
 
 # this makes sure the dependencies for the EGL mode are present; otherwise, the configure scripts
 # automatically and silently fall back to GLX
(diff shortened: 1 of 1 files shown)


meta-chromium/recipes-browser/chromium/chromium-gn.inc
```diff
@@ -109,7 +109,7 @@ BUILD_CC:toolchain-clang = "clang"
 BUILD_CXX:toolchain-clang = "clang++"
 BUILD_LD:toolchain-clang = "clang"
 
-PACKAGECONFIG ??= "upower use-egl"
+PACKAGECONFIG ??= "upower use-egl use-v4l2"
```
otavio, line 112: Personally, I believe that you could add the UDEV rules together with the support, so this works out of box whenever someone enables it.

marex: @otavio those udev rules should be in BSP layer (temporarily, read on ...), since they are SoC specific and this layer cannot predict what the BSP layers are doing with udev rules.

There is already an ongoing work in chromium upstream to use the proper `/dev/videoN` and `/dev/mediaN` nodes , I think that is going to land in chromium 132 or 133 , at which point the udev rules won't be needed at all.

So no, I don't think this layer should be adding any udev rules. The udev rules in the commit message are an example for one specific SoC.

otavio: I agree it usually fits in the BSP layer. Since it's harmless, adding the rules for other processors, too, could be a great idea. We can install them without any conditions. They won't have any impact if used with a different processor so that it wouldn't cause any issues.

Also, I appreciate your argument about using proper device names and how those rules may not be needed shortly. That's another good reason why this might be a better fit here.

kraj: @marex we are now at 132.0.6834.83, I wonder if we should land this PR now

marex: @kraj I think yes , I already rebased it last month and dropped the udev rules part .

kraj: I think having the udev rules alongside would be good.

PR #879, opened 2025-03-25
RFC: Firefox

OldManYellsAtCloud:
I have been maintaining this fork for about 1.5 years now. First I started it only for myself, but I think that by now it might be usable for others too. I'd like to merge it back, mostly for exposure.

This is an RFC first and foremost, to show how it looks now, and more importantly I'd like to know if you would be open for merging, and if yes, what changes you'd like before it would happen.

A couple of notes, in no particular order:
- The layer is of course still largely based on this parent repository, though there are quite some changes
- I maintain recipes for both ESR and latest stable version. So far I have updated to all minor versions (except for 1 Windows-only patch release)
- The layer is compatible (and tested) with all 4 currently supported Yocto releases (Kirkstone, Scarthgap, Styhead, Walnascar/master)
- Both glibc and musl are supported
- Upon PR, all these variations are build- and smoke-tested automatically (Yocto version, FF version, libc version combos, along with arm/aarch64/risc-v/x86-64 arches). There is a separate supporting test repo in my profile: meta-firefox-test. I support non-default PACKAGECONFIGs, but I don't test them regularly.
- The repository has a very few organic users (I estimate no more than 10), but hasn't been exposed to wide audience yet. I have been dogfooding it on aarch64.
- The commit history got reasonable lately, though especially in the beginning it was a bit turbulent. I would prefer not losing the history, but I could definitely squash a lot of things - I think I could shrink the history to about half of the current length - let me know your preference, if you have any.
- The updates became fairly routine since a while, the main thing is updating the crates, and running the CI. Without CI the maintenance takes a good part of an afternoon/month.
- update_recipe_creates job is not used: it generates such a huge list, that's... well, it's measured in megabytes. The create list if generated based on the third-party/rust list, and the content of different Cargo.toml files, and the related dependencies.
- In theory the Firefox tarball contains all sources to build it offline. In practice this is not true. Some crates are missing from the third party folder, and some are misconfigured (e.g. they refer to other crates in the third-party folder with absolute path, which points to a wrong level). That's also one reason to fetch the crates separately.
- When there is a git repo in SRC_URI, the cargo classes add them to the cargo config, as repo-patches. This is disabled, because it can't handle correctly when the repository contains multiple crates. It is handled via patching the relevant Cargo.toml files.
- This layer used many of patches used by Debian. While I kept some, I have dropped most of them - I either felt that they are not relevant, or I found them to be too opinionated.

Some known issues:
- content-sandbox doesn't work with 32-bit arm and with musl (all arches). Currently FF is built without this feature for these combos.
- the "system-av1" PACKAGECONFIG is completely untested for the latest FF (though I imagine that this will change by the time it is merged)

One thing I'd like add: I'd prefer not giving up this project. If this PR gets to a stage that it is accepted, I'd like to remain the primary maintainer of meta-firefox for the time being.

I might add more comments, as they come into my mind.

I guess you have questions and/or comments - go ahead, let me know them.
I know that this is not a small PR, so take your time.


OldManYellsAtCloud:
Marked it as not-draft, as copilot says that github doesn't send notification about draft PRs... but of course it's still a draft. Just wanted send some github emails about it.

otavio:
Hi @OldManYellsAtCloud,

Thank you so much for your fantastic work! We truly appreciate your contributions and understand that updating some patches is a natural part of the process. Merging some commits would be great to keep things neat and tidy. If you have any ideas for improvements to make maintenance a breeze, we would love to hear them!
We’re also excited to invite you to become an official maintainer for Firefox recipes. Your expertise would be an excellent addition!

kraj:
Can this be rebased on top of latest master please ?

OldManYellsAtCloud:
I think that the history is about as good as I can make it.
Alternatively, if you still consider it too unwieldy, I could also squash the whole thing into 1, and keep the history in my fork (which I plan to archive anyway), in case I would need for whatever reason.

I also just explicitly rebased it on the master branch of this repo.

Regarding maintenance, it doesn't take a lot of time currently. Without CI, a regular main version bump takes around an hour of work. A patch release is measured in minutes. (CI run does take a while, and sometimes it shows issues with some configurations, though it doesn't add nowadays significant amount of work - most frequently it's nss/nspr version-bump)

The main hurdle (and point of failure) is dealing with crate updates. Currently the tools provided by Yocto for this task are insufficient for Firefox - would this not be the case it would cut down this work very much. One of these days I might get deeper into that, but that won't happen today...

Otherwise CI takes almost forever - but it's like that. Usually that doesn't need a lot of attention at least, beside restarting it after it times out the first time (it takes currently ~33-35 hours - before I started using it, sometimes I had very broken builds for some configs for weeks without even knowing it, so I'm not really complaining about the time it takes).

Oh yeah, regarding CI builds: it is running on a self-hosted machine of mine, so it doesn't cost money or anything. It is also dockerized, with small assumptions about the host.

PR #880, opened 2025-03-31
Firefox - update to 137.0

OldManYellsAtCloud:
This PR contains the following:

- Update Firefox 128.8.0esr to 128.9.0esr
- Update Firefox 136.0.3 to 137.0

(136.0.4 and 128.8.1esr were intentionally skipped. They addressed a Windows-only vulnerability.)

- Some cleanups
  - Drop vendored nss and nspr recipes, and instead add a patch to allow Firefox's vendored nss to be compiled without crypto support. System nss and nspr can be still used instead of Firefox's own, using a PACKAGECONFIG.
  - Fix url in GitHub workflow file
  - Add "--disable-sandbox" config to arm arch. It went accidentally missing during me squashing the history for the initial merge. Not sure how, but it happened (I also compared this repo with my master branch, and this was the only difference fortunately).

Update after running all build and test jobs:
- Initially I compiled all Walnascar versions with clang20. It seemed to work, except for the ESR version at x86-64 arch, using glibc. That segfaults upon start. For now up to clang19 is supported on Walnascar - will look into that issue and hope to fix with the next release.
- The majority of the time in the past days was spent on mixing and matching Rust. Firefox 137 requires at least Rust 1.82 to compile, however the state of Rust in Yocto has seen better days:
  - Walnascar has reasonably new Rust out of the box, so that works without touching it.
  - Kirkstone uses meta-lts-mixins, but it contains only Rust 1.80.1. For Firefox 137 meta-rust is used, which contains 1.82.
  - Scarthgap uses meta-lts-mixins, but it contains only Rust 1.81.0. meta-rust claims to be compatible with Scarthgap (on the Scartgap branch at least), but unfortunately in practice I couldn't compile the toolchain without fatal errors. I have [forked](https://github.com/OldManYellsAtCloud/meta-lts-mixins) meta-lts-mixins, and ported Rust 1.82 from the master branch of oe-core, which seems to work okay, I used it to build all relevant variations. For Firefox I can support it - otherwise I have some hope still that meta-lts-mixins will be updated in the near future, and I can delete my fork.
  - Styhead comes with Rust 1.79, and unfortunately it seems to be kind of abandoned Rust-wise. I *think* that my fork would work with Styhead with minor modifications, but due to the closeness of its EOL date, I did not try to port it myself. The latest version of FF was not tested with Styhead. If a user finds a new enough Rust recipe that works with Styhead, I have no problem with supporting it as long as it's not EOL.
- Since meta-rust is back, I have removed the hard dependency on meta-lts-rust from layer.conf, and made the RUST_HOST and RUST_TARGET variables configurable from a bbappend/local.conf, because meta-rust and oe-core/meta-lts-mixins use different machine triplets.


OldManYellsAtCloud:
1. Note that these versions haven't been announced, so they are not official yet. It's 99% sure that they will be the final versions, but sometimes it happens that they find some last-last-minute bug, and go for a .1 patch release. Nevertheless, they should be perfect at least for CI run and recipe sanity check. The official release should happen tomorrow, around this time-ish.
2. If you'd prefer these PRs in any different way, please don't hesitate to speak up. (Should I create an issue for the updates? Or would you like 1 commit/PR, or perhaps 1 squashed commit/PR? or anything else that comes to your mind)

3. Regarding CI run:
a. Could you please give me access to this repo, to be able to set up and control GitHub runners and workflows?
b. If no, then could you start setting up a new self hosted runner? If you could send me the details it shows to my email address, I could register my runner here. Also, if you could configure the workflow execution to require approval, that would be very useful (these are fairly long workflows)

otavio:
That is unfortunate, but as usual, that happens when we think that we have everything under control. No worries, take your time. Whenever you have that ready, I think that we can review and merge it.

OldManYellsAtCloud:
Bit of a status update: I managed to solve most of the issues.... but some test jobs have failed. Firefox segfaults on Walnascar with some arches. Investigating that.

OldManYellsAtCloud:
Finally I managed to build and also test all variations that I wanted successfully: https://github.com/OldManYellsAtCloud/meta-browser/actions/runs/14258193663

I have also updated the PR description with some details

meta-firefox/recipes-browser/firefox/firefox/0001-Add-option-to-disable-arm-hw-crypto-engine.patch
```diff
@@ -0,0 +1,86 @@
...
+Subject: [PATCH] Add option to disable arm hw crypto engine
+
+On some older ARM SoC's there is no hardware support for cryptography.
+However the nss librray in Firefox expect this support to be present,
+which fails the build when this is not the case.
+
+To avoid this, introduce a new option that allows disabling building
+for ARM crypto engine when it is required.
+
+Upstream-Status: Pending
```
kraj, line 13: You should submit this patch upstream separately so it gets discussed

kraj (approved):
LGTM

Issue #845, opened 2024-10-31
Chromium 129.0.6668.100 compile fails at the very end during linking with std::bad_alloc error


neenareddi:
Hi,

Chromium 129.0.6668.100 compile fails at the very end during linking with std::bad_alloc error. Built this using meta-clang styhead branch. We tried to build on a machine with high memory but it still failed. 

![image-2024-10-25-16-25-58-156](https://github.com/user-attachments/assets/42931f0e-3017-446e-b2bd-1b73dbc122fb)

Did anyone run into this issue ? Any help would be much appreciated.

MaxIhlenfeldt:
I've been seeing this issue in my builds for the last few releases, but @SoumyaWind was able to build fine.

Maybe it's time to actually submit a bug report to LLVM as suggested in the output. Would it be possible for you to do that?

WujianSun:
On my side also meet the same issue on NXP imx boards

jcalvinowens:
I ran into this today too, the issue is the kernel's limit on the number of memory mappings each process is allowed to create: https://docs.kernel.org/admin-guide/sysctl/vm.html#max-map-count

You can set the limit to a larger number easily, just run this as root:
```
echo 1048576 > /proc/sys/vm/max_map_count
```

To make it permanent, add this to /etc/sysctl.conf:
```
vm.max_map_count = 1048576
```

stu-spp:
I was running into this problem too. Increasing the `vm.max_map_count` to `1048576` fixed it for me. Thank you!!!

MaxIhlenfeldt:
Thanks for finding the fix!

PR #882, opened 2025-04-02
Readme: added section to chromium build for std::bad_alloc

BasketsAlaaf:
This adds a section to the readme, explaining the solution to a build error discussed in #845.

Adding this in the readme makes it easier to be found.


MaxIhlenfeldt:
I think we can be a bit less verbose in the readme. It's enough to list the error message, then the two ways to fix it (temporarily or permanently), and then link to #845 for additional context.

BasketsAlaaf:
I tried to condense it a bit.

meta-chromium/README.md
```diff
@@ -82,6 +82,36 @@ host. clang-native from the meta-clang layer is used to build those binaries.
...
+# echo 1048576 > /proc/sys/vm/max_map_count
+```
+This change will only persist until the system is rebooted. 
+
+2. Permanently Set vm.max_map_count:
+To make this change permanent, you need to modify `/etc/sysctl.conf` and add:
+```bash
+vm.max_map_count=1048576
+```
+A reboot may be required for the new value to get picked up (or run `sysconf -p`)
```
MaxIhlenfeldt, line 113: nit: missing . at the end.

MaxIhlenfeldt (approved):
lgtm once the one nit is addressed :)

MaxIhlenfeldt:
Thanks!

PR #884, opened 2025-04-15
firefox: update 137.0.1 -> 137.0.2

OldManYellsAtCloud:
Patch release, no changelog yet - I expect it to be officially announced in a few hours.


OldManYellsAtCloud:
Let me see is the CI workflow uses the correct branch, or it needs some tweaking

otavio:
The workflow is not being triggered for the pull requests. I think that would be good if you can prepare the fixes for it. So we can merge them and then later rebase this one on top of this.

OldManYellsAtCloud:
Right... I used to have PR-trigger. Now I'm thinking if I switched to manual trigger only because I didn't have the runner here, or if there was any other reason too... but I can't really recall any.

Hold on for a few moments - it has just failed (I forgot to clean the build area properly before running), which is a perfect opportunity for this.

kraj (approved):
LGTM

OldManYellsAtCloud:
The CI runs have finished yesterday successfully - from my end at least, I think this is ready.

PR #892, opened 2025-05-14
firefox: update 138.0.1 -> 138.0.3

OldManYellsAtCloud:
138.0.2 was not released - that's the reason for the version skip.

Also, add "master" branch to the list of workflows to run - though it will run only starting from the next PR.


kraj (approved):

OldManYellsAtCloud:
Hmmm... matchbox-desktop fails to start on master with qemuarm and musl, which fails the workflow. Taking a dive there.

OldManYellsAtCloud:
Well, finding a solution to the mesa problem lasted long enough that meanwhile Mozilla has released a new Firefox version (late Saturday, which is a bit unexpected on its own - but I guess they were in a rush...)

Once the CI run proves to succeed I will close this PR, and open a new one with the latest versions.

PR #894, opened 2025-05-26
Firefox 139

OldManYellsAtCloud:
Firefox 138.0.4 -> 139.0
Firefox 128.10.1esr -> 128.11.0esr

Notes:
- Styhead support is dropped. It is EOL, and unfortunately it has a very old Rust recipe, which couldn't compile the latest Firefox since a couple of versions. If someone wants to use Styhead for some reason, this ESR should still compile successfully, but it's not a too good idea. This PR removes Styhead from the layer compatibility list, and from the GitHub workflows also.
- Added new `PACKAGECONFIG` options, that were just introduced in the Firefox build system. Now it is possible to link against a shared libdrm, gbm and pipewire.
- Added a new, conditional LDFLAGS extension to the latest Firefox version, in case it is compiled against shared av1 libraries. This is a workaround for an issue I encountered during test compilation with this PACKAGECONFIG. It seems to be [reported](https://bugzilla.mozilla.org/show_bug.cgi?id=1951965) to Mozilla already, though it looks unfortunately they marked is a duplicate of a similar (but still, different) issue... I think I will need to revisit this deeper.


OldManYellsAtCloud:
From my end this is ready for this version.

Issue #821, opened 2024-06-03
Is it possible to install  firefox with wayland and without x11 lib


yuchinchenTW:
I have encounter same issue https://github.com/OSSystems/meta-browser/issues/533
it seems like this problem exist couple years ago 

is there a way to install firefox on wayland and without x11 nowadays?

PR #896, opened 2025-06-21
Run firefox without X11

Gerry546:
This commit should remove the dependencies on needing X11 to build. So far I haven't had trouble building a wayland only system where I can compile and run a firefox browser.

Related to #821


OldManYellsAtCloud:
Thanks for this, it looks good at the first sight. Currently doing some work on the CI machine, but I should be able to run the workflow in about a day-ish.

OldManYellsAtCloud:
In about 2 hours from writing this comment I will have to cancel the workflow temporarily - there will be a short-scheduled power outage at my place, which supposed to take an hour or two. Will continue with the remainder once electricity is back.

PR #903, opened 2025-07-22
firefox: update to 140ESR and 141, whinlatter 

OldManYellsAtCloud:
This is a bit busier PR than usual:

1. Added whinlatter compatibility.
- clang-layer is now only required for pre-whinlatter versions.
- This brought to the surface that wasi-sdk is not cmake4 compatible. For now I have added a patch to alleviate this issue. I have started to update wasi-sdk a few times, this should give more motivation to bring it through.
- It also brought up a small difference between clang in oe-core and meta-clang: meta-clang has `-Qunused-arguments` added to CCARGS by default, which is missing from oe-core (and it fails the configuration for 32-bit arm targets). I have added a small patch to compensate this in Firefox. Still checking if there is anything that is affected in meta-oe and oe-core by this, so it might worth to submit a patch to oe-core, or we should just live with this.
- Similar to what was discussed in https://github.com/OSSystems/meta-browser/pull/897, I have added an include in layer.conf that sets the required `S` for some recipes that need different one in whinlatter and pre-whinlatter. Initially I planned to include it conditionally, depending on the value of LAYERSERIES_CORENAMES - however it seems that require and include directives are processed before this variable would get a value, so I couldn't rely on this, it is always empty at the time when the includes are processed. So the folders are now set both pre- and post-whinlatter, and they are set based on the current version.
- (plus whatever comes up during the workflow run)

2. Dropped the v128 ESR series, and switched to the new ESR series, v140.

3. Updated the regular, latest release from 140 to 141.
- Since the ESR and latest versions are so close to each other, there is a lot of overlapping between the recipes, so I have unified the recipes and the patches trying to avoid duplications.

@jcalvinowens Unfortunately I am running out of storage space, and unable to keep the old master build artifacts further around, had to do the upgrade. I have added you as a co-author to the whinlatter upgrade. Hope you don't mind.


jcalvinowens:
> Hope you don't mind.

No problem at all, thanks for finishing it!

PR #924, opened 2025-09-18
Firefox: upgrade 142.0.1 -> 143.0

OldManYellsAtCloud:
This PR contains a regular Firefox update.

Two notes I'd like to add:

1. The latest Firefox contains an update of the cc crate, and it brought some changes when it comes to cross-compiling. After a recent change, it reads more environment variables than it used to to gather the CFLAGS/CXXFLAGS variables. This causes problems when some build-only dependencies are compiled by rust, because it mixes the CFLAGS (target flags) with HOST_CFLAGS (native flags). To avoid the issue, added a new patch: 0001-cc-crate-skip-HOST_CFLAGS-and-HOST_CXXFLAGS-env-vars.patch. This patch is only suitable for Firefox target compilation. Relevant upstream issue: https://github.com/rust-lang/cc-rs/issues/1469
2. I have removed the includes for crates. When I first created these includes, the Firefox tarball had a number of missing crates causing build failures. But I haven't kept checking if it has improved since. Now, while investigating the above build issue, I noticed that it seems to build fine with the crates from the tarball also, so let's just use that.


OldManYellsAtCloud:
Mozilla has just released Firefox 143.0.1, however it contains only a fix for a Windows-only crash (some Trend Micro DLL injection makes FF unhappy). I plan to skip that version.

OldManYellsAtCloud:
Yes, it timed out (if it doesn't finish in 24 hours, it cancels the non-executed jobs). The jobs usually take at least 30 hours (but if a lot of packages need to be rebuilt then more), so it always times out at least once.

Meanwhile I had also noticed some issues with the recipes which I have just corrected - I plan to push it in minutes, and restart the CI jobs.

PR #925, opened 2025-09-23
chromium: Remove gitfiles present in chromium tarball

jcalvinowens:
The chromium release tarballs contain several ancillary gitfiles like .gitignore. These files confuse devtool, and prevent its use for managing chromium patches until they are manually removed.

Since they serve no purpose, append a do_unpack stub which removes them.

Arguably, devtool should ignore ancillary gitfiles if there is no actual .git directory present: I will follow up on that upstream as well.


jcalvinowens:
The problem doesn't reproduce anymore despite the files still being there. I think https://lore.kernel.org/all/20250702105658.3862855-1-mac@mcrowe.com/ inadvertently fixed this.

Issue #875, opened 2025-03-09
Can't build meta-chromium with python3.13


jcalvinowens:
This happens because the disttools module is no more, see https://peps.python.org/pep-0632/
```
DEBUG: Executing python function do_check_llvm_version
ERROR: Error executing a python function in exec_func_python() autogenerated:

The stack trace of python calls that resulted in this exception/failure was:
File: 'exec_func_python() autogenerated', lineno: 2, function: <module>
     0001:
 *** 0002:do_check_llvm_version(d)
     0003:
File: '/home/calvinow/git/meta-roosevelt/build/../meta-browser/meta-chromium/recipes-browser/chromium/chromium-gn.inc', lineno: 400, function: do_check_llvm_version
     0396:# Check the LLVMVERSION defined in the meta-clang layer. Given Chromium is
     0397:# developed using new C++ features, the LLVMVERSION has to be >= 14. Otherwise,
     0398:# it is not guaranteed that Chromium will compile.
     0399:python do_check_llvm_version () {
 *** 0400:  from distutils.version import LooseVersion
     0401:
     0402:  min_llvm_version = "14.0.0"
     0403:  llvm_version = d.getVar('LLVMVERSION', False)
     0404:  if not llvm_version:
Exception: ModuleNotFoundError: No module named 'distutils'

DEBUG: Python function do_check_llvm_version finished
```

I haven't had time to spin a PR to fix this yet, but will soon if somebody doesn't beat me to it.

jcalvinowens:
Sorry for dropping this, it's since been fixed.

PR #927, opened 2025-09-24
chromium: Remove defunct LLVM version check logic

jcalvinowens:
The new core LLVM support lacks an equivalent of LLVMVERSION as exposed by meta-clang, resulting in this benign WARNING:

    WARNING: mc:house:chromium-ozone-wayland-140.0.7339.136-r0 do_check_llvm_version: Unable to determine LLVM version. Chromium may not be compiled successfully if the LLVM version is below 14.0.0.

LLVM-14 was released nearly four years ago. Upstream chromium's current policy is summarized here:

    https://chromium.googlesource.com/chromium/src/+/main/docs/toolchain_support.md

...and quoting the releavant portion:

    Even within clang/libc++, we often rely on recent bugfixes and
    feature additions, so generally only the most recent LLVM version
    is supported.

Silence the warning by dropping the bitrotted version enforement logic.

### DIFF
--- a/meta-chromium/recipes-browser/chromium/chromium-gn.inc
+++ b/meta-chromium/recipes-browser/chromium/chromium-gn.inc
@@ -420,26 +420,6 @@ python do_create_v8_qemu_wrapper () {
 do_create_v8_qemu_wrapper[dirs] = "${B}"
 addtask create_v8_qemu_wrapper after do_patch before do_configure
 
-# Check the LLVMVERSION defined in the meta-clang layer. Given Chromium is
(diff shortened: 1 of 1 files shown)


jcalvinowens:
The warning shows up after https://github.com/OSSystems/meta-browser/pull/923 but this change doesn't actually depend on that PR.

If you'd rather keep the version check in some form just say the word and I'll do that instead.

jcalvinowens:
This also fixes https://github.com/OSSystems/meta-browser/issues/875

PR #934, opened 2025-10-18
firefox: upgrade to 145

OldManYellsAtCloud:
This the regular FF update, however it is not ready yet - version 144 requires a newer wasi sdk, I really can't postpone that upgrade any further. This will become a real PR once I succeed with that - will try to spend some time on that in the coming day.


OldManYellsAtCloud:
I couldn't get still make wasi-sdk work in a way that doesn't take forever to build. It builds its own clang, which of course takes forever. But making it work with the already existing clang just keeps beating me up. For now I have disabled the wasi-sdk part in the latest Firefox recipe (enabled `disable-sandboxed-libraries` PACKAGECONFIG by default)

OldManYellsAtCloud:
I believe I managed to update wasi-sdk successfully. 

The Github workflow has just started, but did a couple of tests locally on all branches, and the basics look correct. I don't expect big issues during the I run (\me knocking on wood).

Unfortunately I couldn't make my original goal work (to use the main clang for WebAssembly also), mostly due to compiler-rt: wasi-sdk wants its own compiler-rt, and I'm not sure if that is possible at all without having its own sysroot. Maybe I will take another look at it on a rainy day. For now it will build its own toolchain, just as until now.

A couple of further notes about this update:
- When using wasi-sdk as a dependency, at least clang 18 is required. This is only a little bit notable, since it is available for Kirkstone also in meta-clang (and other branches have this or newer by default). When not using wasi-sdk, clang 14 should still work. Currently I switched to test only with clang 18 on Kirkstone.
- This is the last main Firefox version that is tested with Walnascar. Starting with Firefox 146 I plan to drop testing on Walnascar - though I plan to keep it in the compatibility list and give user support as long as it makes sense.
- Since Whinlatter is essentially still Master, I'm not testing it specifically for this version. Starting with Firefox 146 Whinlatter will be part of the test suite explicitly.

Will ping back when/if the workflow finishes successfully. Meanwhile, please feel free to review/comment.

OldManYellsAtCloud:
Well, the CI run was a bit more painful than expected, but not because of this update, just some missing files here and there... but it is done now: https://github.com/OSSystems/meta-browser/actions/runs/19741494114

From my end it is as good as it gets, and it is ready.

PR #948, opened 2025-12-10
Firefox 146

OldManYellsAtCloud:
Regular update, so far mostly eventless as far as updates go. Walnascar has been dropped from the CI config, and Whinlatter was added.


kraj (approved):
LGTM

PR #968, opened 2026-03-26
meta-chromium: Support only wrynose

zboszor:
With the gn recipe addendum removed, Chromium 145/146 can't be built from master when coupled with the openembedded-core whinlatter branch.


MaxIhlenfeldt (approved):
lgtm, thanks!

Issue #974, opened 2026-04-14
Persistent build error about missing libclang_rt.builtins.a


tronje:
Hi, I'm getting the same error as in #947, and #962 does not fix it for me. I'm using the current `master` state of `meta-browser`. Here's the error again:

```
ERROR: chromium-x11-146.0.7680.177-r0 do_compile: Execution of '/workdir/build/tmp/work/corei7-64-poky-linux/chromium-x11/146.0.7680.177/temp/run.do_compile.1031968' failed with exit code 1
ERROR: Logfile of failure stored in: /workdir/build/tmp/work/corei7-64-poky-linux/chromium-x11/146.0.7680.177/temp/log.do_compile.1031968
Log data follows:
| DEBUG: Executing shell function do_compile
| ninja: error: '../../../../recipe-sysroot-native/usr/lib/clang/latest/lib/x86_64-unknown-linux-gnu/libclang_rt.builtins.a', needed by 'yocto_native_for_rust_host_build_tools/obj/build/rust/allocator/liballoc_error_handler_impl.a', missing and no known rule to make it
| WARNING: /workdir/build/tmp/work/corei7-64-poky-linux/chromium-x11/146.0.7680.177/temp/run.do_compile.1031968:148 exit 1 from 'ninja -v -j 20 chrome chrome_sandbox chromedriver.unstripped'
| WARNING: Backtrace (BB generated script):
| 	#1: do_compile, /workdir/build/tmp/work/corei7-64-poky-linux/chromium-x11/146.0.7680.177/temp/run.do_compile.1031968, line 148
| 	#2: main, /workdir/build/tmp/work/corei7-64-poky-linux/chromium-x11/146.0.7680.177/temp/run.do_compile.1031968, line 152
ERROR: Task (/workdir/layers/openembedded-core/../meta-browser/meta-chromium/recipes-browser/chromium/chromium-x11_146.0.7680.177.bb:do_compile) failed with exit code '1'
```

Any help? I'm quite puzzled by this.

I've tried with various versions of the other common layers (meta-yocto, openembedded-core, meta-openembedded, etc.), including their latest master states. I've also tried with and without `meta-clang`. `meta-clang` generally puzzles me a bit, because it doesn't actually provide a recipe for clang anymore.

Thanks very much for all your work!

MaxIhlenfeldt:
[...]

PR #975, opened 2026-04-16
chromium: Restore patch for fixing path to libclang_rt.builtins.a

jcalvinowens:
This patch is still required on some systems to avoid this error:

    ninja: error: '.../libclang_rt.builtins.a', needed by '.../liballoc_error_handler_impl.a', missing and no known rule to make it

See https://github.com/OSSystems/meta-browser/issues/974


meta-chromium/recipes-browser/chromium/files/0008-Use-the-correct-path-to-libclang_rt.builtins.a.patch
```diff
@@ -0,0 +1,63 @@
...
+ 1 file changed, 9 insertions(+), 8 deletions(-)
+
+diff --git a/build/config/clang/BUILD.gn b/build/config/clang/BUILD.gn
+index e05032c0c66f1..6d35b7e1d8412 100644
+--- a/build/config/clang/BUILD.gn
++++ b/build/config/clang/BUILD.gn
+@@ -185,22 +185,23 @@ template("clang_lib") {
+       } else if (is_apple) {
+         _dir = "darwin"
+       } else if (is_linux || is_chromeos) {
```
kraj, line 32: Is there platform define available for `openembedded` or `yocto`

jcalvinowens:
I think I figured this out, with https://github.com/OSSystems/meta-browser/pull/976 I don't need the 0008 patch and we can just drop this I think.

PR #976, opened 2026-04-17
chromium: Fix build with non-default TARGET_VENDOR

jcalvinowens:
The sed invocation assumes TARGET_VENDOR is "-oe", which causes this error for yocto configurations which set TARGET_VENDOR or HOST_VENDOR:

    ninja: error: '.../libclang_rt.builtins.a', needed by '.../liballoc_error_handler_impl.a', missing and no known rule to make it

Fix by using "${TARGET_VENDOR}-" and "${HOST_VENDOR}-", not "-oe-".

See https://github.com/OSSystems/meta-browser/issues/974


zboszor (approved):

kraj (approved):
LGTM

PR #981, opened 2026-05-11
firefox: upgrade 150.0.1 -> 150.0.3

OldManYellsAtCloud:
Bug fix releases:
https://www.firefox.com/firefox/150.0.2/releasenotes/
https://www.firefox.com/firefox/150.0.3/releasenotes/
https://www.firefox.com/firefox/140.10.2/releasenotes/

(While running the tests for 150.0.2, Mozilla released 150.0.3 also, so I added it to this PR)

Also, add recipe for wasm-tools. This is a dependency for wasi-sysroot - until now it was downloaded silently, during do_compile.

The network cannot be always disabled by bitbake after do_fetch, because it requires some specific capabilities. The docker image that I use for testing is missing this, and apparently network access is always available in that image, this is how it was missed. Will try to look into that, if I can make network disabling work in docker.


OldManYellsAtCloud:
CI passed. From my end this is ready.

PR #990, opened 2026-05-29
firefox: remove hard clang dependency

OldManYellsAtCloud:
TBH I don't remember anymore why I made initially Firefox require clang, unfortunately it was too long time ago.
But today I was curious if gcc works - and it does, it required only a minor patch.

This PR removes the hardcoded clang dependency from the Firefox recipe.

After this clang is only required if Firefox is built with sandboxed libraries enabled (this is the default, and can be changed with a `PACKAGECONFIG`. This is due to wasi-sdk having a hard dependency on clang)

When built without sandboxed libraries, clang can be avoided completely (at least from Firefox's point of view).

So far I tested successfully Scarthgap with glibc on x86-64, with both versions of Firefox. The rest of the test is left to the CI.

### DIFF
--- a/meta-firefox/classes/mozilla.bbclass
+++ b/meta-firefox/classes/mozilla.bbclass
@@ -36,6 +36,8 @@ export WASM_CXX = "${WASI_SYSROOT}/bin/clang++ -target wasm32-wasip1 "
 
 export BUILD_VERBOSE_LOG = "1"
 
+export MOZ_MAKE_FLAGS = "${PARALLEL_MAKE}"
+
 RUST_HOST ?= "${RUST_BUILD_SYS}"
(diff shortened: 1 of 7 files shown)


OldManYellsAtCloud:
This PR is on hold for a bit. When compiled with gcc, 32-bit arm builds of the latest FF crash upon the very first start. 

Looks like some random memory corruption somewhere (which manifests with a different error and different gdb backtrace each run :) ), and I'm afraid I will have to bisect it to get any further, which will take some time.

OldManYellsAtCloud:
Spending a few hours debugging this before looking up the documentation helped me understand why I made it clang-only: Firefox has only tier3 (community-based) support for gcc for anything other than x86-64. I hope I won't forget it for a while.

PR #915, opened 2025-09-01
Electron37, Chromium138 and GitHub actions for automated builds/testing

caneraltinbasak:
This pull request introduces several new GitHub Actions workflows to automate build, test, and cleanup processes for Chromium and Electron on AWS EC2, and updates the Yocto layer and Chromium build configuration for improved compatibility and hardware support. Key changes include the addition of EC2-based CI workflows, a cleanup workflow for EFS cache, support for new Yocto releases.

**CI/CD Workflow Additions and Improvements:**

* Added `.github/workflows/bs_meta_browser_ci_ec2.yml` to orchestrate EC2-based build and test jobs for the meta-browser project, including runner lifecycle management and parameterized builds for different architectures, browsers, and Yocto versions.
* Introduced `.github/workflows/chromium.yml` and `.github/workflows/electron.yml` to enable matrix builds and smoke tests for Chromium and Electron, supporting both AWS EC2 and local runners, with flexible configuration via workflow dispatch and PR triggers. [[1]](diffhunk://#diff-8f5503c36aa7e90a80695b902b7758301560c1c8c3e8e961455dbddad3a7b0f2R1-R102) [[2]](diffhunk://#diff-e3f2d8bfe2b2f0cc1deb963d1584f4f7cf6fff5111a0fd4c1d90904e1d7c93ffR1-R102)
* Added `.github/workflows/cleanup-efs-cache.yml` to automate cleanup of EFS cache directories (`downloads` and `sstate-cache`) using a temporary EC2 runner, improving disk space management for Yocto builds.

**Yocto Layer and Build System Updates:**
* Updated Chromium to Chromium138
* Added a recipe for building Electron(Electron37)
* Updated `LAYERSERIES_COMPAT_chromium-browser-layer` in `meta-chromium/conf/layer.conf` to include compatibility with the new Yocto release `whinlatter`.
* In `meta-chromium/recipes-browser/chromium/chromium-gn.inc`:
  - Replaced several patches with new ones to fix Node.js version checks and ARM V4L2 compiler errors, and enabled stateless V4L2 M2M video decoding support via `use-v4l2` PACKAGECONFIG. [[1]](diffhunk://#diff-436bebc7f6c04ff6040b2273ced509fa01ef9453a4fe30c569445ee21e37e364L28-R31) [[2]](diffhunk://#diff-436bebc7f6c04ff6040b2273ced509fa01ef9453a4fe30c569445ee21e37e364R147-R150)
[...]

PR #994, opened 2026-06-02
chore: BCN 21398 update github actions

DPopov-BS:
BCN 21398 update github actions


MaxIhlenfeldt:
I think I'm missing context on this, (why) do we want CI in this repo?

OldManYellsAtCloud:
> I think I'm missing context on this, (why) do we want CI in this repo?

I suspect this is a followup of https://github.com/OSSystems/meta-browser/pull/915 that had a short discussion about this.

PR #1002, opened 2026-06-14
firefox: upgrade 151.0.3 -> 151.0.4

OldManYellsAtCloud:
Release notes:
https://www.firefox.com/firefox/151.0.4/releasenotes/


OldManYellsAtCloud:
This is ready from my end, at least for 151.0.4.

There is however now version 152, so I guess this PR could be left for a bit, and I can just update this PR to the new version later this week.

otavio:
I believe we can merge this as it is, and when you have time, you can start a new one.

PR #1004, opened 2026-06-20
firefox 152.0, test changes

OldManYellsAtCloud:
This PR contains 2 major-ish changes - but would like to avoid running the CI multiple times on the same recipes.

1. Firefox upgrade to latest version. This is nothing too unusual.
2. I have reworked the CI workflows a bit.

I'd like to write a couple of words about the latter change.
Until now testing happened for all stable Yocto releases (+master), with 4 CPU platforms, 2 libc implementations, executed only on x11.
This already takes quite a while to finish, and adding even a single new dimension to the test matrix doubles the execution time.

But I would like to test a bit more aspects than what's done currently. This change still builds and tests both Firefox recipes for all Yocto releases (+master), with 4 CPU platforms. However for each run the libc implementation (glibc or musl) and the display system (x11 or wayland) are randomly selected.

A drawback is that some (random) previously tested configurations are not tested in each CI run.
As an expected benefit the test coverage will be higher just after a few run, I hope this will make the layer more robust. Also, the test execution time should be somewhat shorter (at least from the next upgrade, after I managed to weed out the bugs from this PR that I missed during preparation). It should be also easier to add new tested aspects with this random method without unacceptably big execution time penalty.

----
One extra small change in this PR: the `PARALLEL_MAKE` variable was ignored so far during build. This should be now fixed.


kraj (approved):
LGTM

OldManYellsAtCloud:
It seems that weston fails to start up on wrynose with 32-bit arm when compiled with musl :(, which puts testing on hold for a bit until I (or anyone else) figures out what's wrong

OldManYellsAtCloud:
First I bisected oe-core to find a problem with musl. Then once I fixed that locally, it was still broken, so i bisected oe-core once again this time with my patch, to find that it was broken later by a mesa upgrade also.

For now I give up - I will blacklist arm+musl+wayland combo on master and wrynose in the CI.

I have filed a bug at YP: https://bugzilla.yoctoproject.org/show_bug.cgi?id=16334

OldManYellsAtCloud:
One last note: in the workflow.yml file the docker image got a new volume path. That's a temporary measure: during debugging my sstate folder outgrew its drive, and had to move it to a bigger one. I intend to trim that folder and reverse that change with the next upgrade.

Also, since Mozilla has meanwhile released 152.0.3, I bumped the recipe to that version.

But all in all, this is ready from my end.

OldManYellsAtCloud:
Since yesterday 152.0.4 was also released, bumped this PR and ran the CI on that version too, successfully.

otavio (approved):

PR #1016, opened 2026-07-21
Firefox 153

OldManYellsAtCloud:
Upgrade to latest Firefox releases.

Release notes:
https://www.firefox.com/firefox/140.13.0/releasenotes/
https://www.firefox.com/firefox/153.0/releasenotes/

While touching it, also upgrade wasi-sdk to the latest version (also bump its dependencies).

There would be one noteworthy thing: Firefox 153 requires at least Clang 19 to compile, which is a bump from Clang 17. In general this is not a big issue: Wrynose and Master already ships a newer version, and meta-clang has a branch for Clang 20 to be used with Scarthgap. But Scarthgap users might need to update meta-clang, if they haven't done so already.
This doesn't affect the current ESR version, which should still compile with Clang 17.
(As a sidenote, v153 will be the next ESR, so this will quickly become the baseline in this layer in a release or two)

FF 153 also requires a newer cbindgen than what's available in meta-oe (it requires 0.29.4), so that vendored recipe is also updated.


OldManYellsAtCloud:
This PR is ready from my end.

One small sidenote: I marked the new `riscv-undefined-hwprobe.patch` as Inappropriate, but later I decided to report the issue and submit the patch upstream. Will update the Upstream-Status with the next PR - didn't want to trigger a new CI run for this.

otavio (approved):

PR #1017, opened 2026-07-29
firefox: upgrade 153.0 -> 153.0.1

OldManYellsAtCloud:
Release notes: https://www.firefox.com/firefox/153.0.1/releasenotes/

Drop riscv-undefined-hwprobe.patch, and add detect-new-riscv64-extensions-on-old-kernel.patch
These patches solve the same problem, but the latter one is a backported fix from Firefox, instead of a homebrew fix.


OldManYellsAtCloud:
Ready from my side, CI succeeded.

PR #1025, opened 2026-09-11
chromium: Handle current wrapper template placeholders

ashie:
Chromium's wrapper template uses a lowercase `channel` placeholder and an unterminated `PROGNAME` placeholder in current releases. Replace these forms during installation while retaining the substitutions for older templates.


ashie:
The wrapper template was changed to use lowercase and unterminated
placeholders by an upstream Chromium change during the Chromium 145
development cycle. We confirmed the new format in the Chromium 147
and 150 release tarballs.

https://github.com/chromium/chromium/commit/db3d1a330291

ashie:
Closing this PR after further investigation.

0003-wrapper-extra-flags.patch changes `@@PROGNAME` back to
`@@PROGNAME@@`, so the existing substitution in do_install() correctly
replaces it with chromium-bin. The startup failure I observed was
caused by a modification to that patch in my local Chromium 150
backport, rather than a bug in master.

https://github.com/clear-code/meta-browser/tree/scarthgap-chromium-150

The lowercase `@@channel` placeholder remains unsubstituted, but that
is a separate issue from the startup failure described here.

Sorry for the confusion.

PR #1026, opened 2026-09-12
chromium: Adapt wrapper substitutions to the current template

ashie:
The extra flags patch currently changes `@@PROGNAME` to `@@PROGNAME@@`, which allows the existing `do_install` substitution to resolve the executable path. However, the lowercase `@@channel` placeholder remains unsubstituted.

Preserve `@@PROGNAME` in the extra flags patch and handle both current placeholders in `do_install`.

Supersedes #1025 with a revised approach.


meta-chromium/recipes-browser/chromium/files/0003-wrapper-extra-flags.patch
```diff
@@ -25,4 +25,4 @@ index aaa46bf..7d8c8dd 100755
 +
  # Note: exec -a below is a bashism.
 -exec -a "$0" "$HERE/@@PROGNAME" "$@"
-+exec -a "$0" "$HERE/@@PROGNAME@@" ${CHROME_EXTRA_ARGS} "$@"
++exec -a "$0" "$HERE/@@PROGNAME" ${CHROME_EXTRA_ARGS} "$@"
```
MaxIhlenfeldt, line 29: Is `CHROME_EXTRA_ARGS` doing something? We declare it empty and never change it. It looks to me like we can just remove both its declaration and its use in this line.

MaxIhlenfeldt (changes requested):
Please mention https://crrev.com/c/7253206 in the commit message as the upstream change that made this necessary.

meta-chromium/recipes-browser/chromium/files/0003-wrapper-extra-flags.patch
ashie: It seems that adding it is the purpose of this patch.
`CHROME_EXTRA_ARGS` is a placeholder, it's replaced by `sed` command in chromium-gn.inc.

https://github.com/OSSystems/meta-browser/blob/9e14b5686c42c558e6923d4b2c41b58790033c42/meta-chromium/recipes-browser/chromium/chromium-gn.inc#L596-L598

ashie:
> Please mention https://crrev.com/c/7253206 in the commit message as the upstream change that made this necessary.

Thanks for your review! I've added it.

meta-chromium/recipes-browser/chromium/files/0003-wrapper-extra-flags.patch
MaxIhlenfeldt: Ah, thanks. In that case, please add this before the `CHROME_EXTRA_ARGS` declaration in line 24 of the patch:
```
// Modified in do_install() in chromium-gn.inc.
```

ashie: Thanks, I'll add it
```diff
++# Modified in do_install() in chromium-gn.inc.
 +CHROME_EXTRA_ARGS=""
 +
```

ashie: Done.

MaxIhlenfeldt: Thanks!

MaxIhlenfeldt (approved):
lgtm, thanks!

MaxIhlenfeldt:
Could you please send cherry-picks of this PR to the `wrynose` and `whinlatter` branches as well? That would be very helpful.

PR #1027, opened 2026-09-17
chromium: Adapt wrapper substitutions to the current template (whinlatter)

ashie:
The extra flags patch currently changes `@@PROGNAME` to `@@PROGNAME@@`, which allows the existing do_install substitution to resolve the executable path. However, the lowercase `@@channel` placeholder remains unsubstituted.

The upstream change at https://crrev.com/c/7253206 updated the wrapper template placeholders, which made this adjustment necessary.

Preserve `@@PROGNAME` in the extra flags patch and handle both current placeholders in do_install.

Backport of #1026


MaxIhlenfeldt (approved):
lgtm, thanks!

PR #1028, opened 2026-09-17
chromium: Adapt wrapper substitutions to the current template (wrynose)

ashie:
The extra flags patch currently changes `@@PROGNAME` to `@@PROGNAME@@`, which allows the existing do_install substitution to resolve the executable path. However, the lowercase `@@channel` placeholder remains unsubstituted.

The upstream change at https://crrev.com/c/7253206 updated the wrapper template placeholders, which made this adjustment necessary.

Preserve `@@PROGNAME` in the extra flags patch and handle both current placeholders in do_install.

Backport of #1026


MaxIhlenfeldt (approved):
lgtm, thanks!

AndreasUfert:
Am I the only one ending up with a broken chromium-wrapper in my image? The last line missing?


```
$ grep meta-chromium /etc/buildinfo 
meta-chromium     = wrynose:6395e0c3946c189cd02d636bd120c7439ae8c244
$ ls -al /usr/bin/chromium          
lrwxrwxrwx    1 root     root            34 Mar  9  2018 /usr/bin/chromium -> /usr/lib/chromium/chromium-wrapper
$ tail -15 /usr/lib/chromium/chromium-wrapper
export CHROME_VERSION_EXTRA="stable"

# We don't want bug-buddy intercepting our crashes. http://crbug.com/41012584
export GNOME_DISABLE_CRASH_DIALOG=SET_BY_GOOGLE_CHROME

# Sanitize std{in,out,err} because they'll be shared with untrusted child
# processes (http://crbug.com/40079595).
exec < /dev/null
exec > >(exec cat)
exec 2> >(exec cat >&2)

# Modified in do_install() in chromium-gn.inc.
CHROME_EXTRA_ARGS="         --use-angle=gles-egl         --kiosk --no-first-run --incognito                    --ozone-platform=wayland"

# Note: exec -a below is a bashism.
$ 
```

AndreasUfert:
Yes, I think a bug was indeed introduced with commit https://github.com/OSSystems/meta-browser/pull/1028/commits/4aa7c8a056185bf588a1baf7e00bf9df119a17b0:

The patch states (now just as before this commit) "3 insertions(+), 1 deletion(-)", even though 4 insertions are actually listed. This, combined with "@@ -36,5 +36,7 @@", ultimately results in the wrapper losing its final line—the one that was supposed to launch the browser.

@ashie @MaxIhlenfeldt Do I need to open an issue for that?

ashie:
Sorry, this was caused by me addressing the review comment hastily. I’ll fix it.

ashie:
I've created a follow up PR: #1031

